PR #308, opened 2017-09-15
add possibility to setup format for date field

oRastor:
e.g. "YYYY-MM-DD" for date only, "HH:mm" for time only


tabuna:
why closed?

oRastor:
Automatically closed after master branch reset. I made the new pull request with not a master branch.

PR #656, opened 2018-05-27
Make views overridable

jebog:
Fixes #

## Proposed Changes

  - Make views overridable
  - make use of `php artisan publish`
  - copy `resources/views` from package to `resources/views/vendor/orchid/dashboard`


tabuna:
I'll be glad to take it into the develop branch

PR #836, opened 2019-04-17
Update menu_controller.js

algha:
The page at 'https://www.xxxxxx.com/dashboard/press/menu' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'https://www.xxxxxx.com/dashboard/press/menu'. This request has been blocked; the content must be served over HTTPS.

to fix this problem, we have to remove '/' from last path of the url.
ps: https://stackoverflow.com/a/47478891

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/resources/js/controllers/components/menu_controller.js
+++ b/resources/js/controllers/components/menu_controller.js
@@ -278,10 +278,8 @@ export default class extends Controller {
      * @returns {*}
      */
     getUri(path = '') {
-        if (path == '') {
-          return platform.prefix(`/press/menu`);
-        }
-        return platform.prefix(`/press/menu/${path}`);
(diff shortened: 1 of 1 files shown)


resources/js/controllers/components/menu_controller.js
```diff
@@ -278,6 +278,9 @@ export default class extends Controller {
      * @returns {*}
      */
     getUri(path = '') {
+        if (path == '') {
```
tabuna: What do you think of such a record?
```javascript
path = path === '' ? `/press/menu` : `/press/menu/${path}`;
return platform.prefix(path);
```

algha: of course, that is cool.
I will change.

PR #837, opened 2019-04-17
Update input_controller.js

algha:
if mask is null or empty, cant not type japanese character.
in this edit, solved typing japanese character.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/resources/js/controllers/fields/input_controller.js
+++ b/resources/js/controllers/fields/input_controller.js
@@ -11,10 +11,9 @@ export default class extends Controller {
 
         try {
             mask = JSON.parse(mask);
+            Inputmask(mask).mask(element);
         } catch (e) {
             // default
         }
-
-        Inputmask(mask).mask(element);
     }
 }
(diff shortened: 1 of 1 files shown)


tabuna:
I tried to reproduce, but it does not work for me. Any language was inserted without any problems. 
But I wrote only in English and Russian, the rest was inserted by copying. 
Does this happen at the moment of pressing?
Снимок экрана от 2019-04-17 10 40 28

tabuna:
I understand that it may be difficult to explain and reproduce.
 But the change is not big, let's take it

resources/js/controllers/fields/input_controller.js
```diff
@@ -14,7 +14,8 @@ export default class extends Controller {
         } catch (e) {
             // default
         }
-
-        Inputmask(mask).mask(element);
+        if (mask != '') {
```
tabuna: As far as I remember, the empty string returns `false`.
```php
if (mask) {
    Inputmask(mask).mask(element);
}
```

algha: ok, thanks.

PR #847, opened 2019-04-25
Adding support of the html in the relationship field

Nks:
image

Sometimes we need to use images/icons/custom html code in relations. Select2 support it without pain but with small edits.


Nks:
Additionally added HTML support in the help attributes for the fields.

image

tabuna:
This may affect security, what will happen if you write javascript code? (Relationship field)

Nks:
> This may affect security, what will happen if you write javascript code? (Relationship field)

So, of course this may cause some problems with security, but at the same time - this is a backend package and everything my point that we can't prevent developers to use the HTML for the help attribute because in many situations developer want to give for users more information about "how they can use this package" without creating a lot of problems. Of course if somebody would like to create javascript inside help attribute - why not? But who will do that? :)

tabuna:
I am more afraid of unexpected results.

For example, a field of `Relationship` that are displayed by usernames for selection, 
but some malicious user will specify a code as their own javascript code.

This change will also affect the `Relation` field in the future.
Perhaps we should do a separate method for this to show that the data will not be sanitized.

```php
Relationship::make('gregre')
    ->allowHtml();

Relation::make('gregre')
    ->allowHtml();
```

What do you think about that?

Issue #862, opened 2019-05-13
Required keyDiff for the Metric layout


Nks:
### Expected Behavior
keyDiff is not required and not showing when it is not exists

### Actual Behavior
keyDiff is required and call an exception when it is not set

### Steps to Reproduce the Problem

  1. Add metric layout and data without keyDiff value
  1. Open the screen with the used layout

### Specifications

  - ORCHID Version: master

Nks:
Assign this on me, I'll fix

PR #863, opened 2019-05-13
refs: #862 Required keyDiff for the Metric layout

Nks:


resources/views/container/layouts/metric.blade.php
```diff
@@ -8,12 +8,16 @@
                 <small class="text-muted block mb-1">{{ __($key) }}</small>
                 <p class="h4 mb-1 text-black font-thin">{{ $metric['keyValue'] }}</p>
 
-                @if((float)$metric['keyDiff'] < 0)
-                    <small class="text-danger">{{ $metric['keyDiff'] }} % <i class="icon-arrow-down"></i></small>
-                @else
-                    <small class="text-success">{{ $metric['keyDiff'] }} % <i class="icon-arrow-up"></i></small>
+                @if(isset($metric['keyDiff']))
```
tabuna: it can be written easier
‘’’php
@isset($records)
 // $records is defined and is not null...
@endisset
‘’’

resources/views/container/layouts/metric.blade.php
```diff
@@ -8,12 +8,16 @@
...
+                    @if((float)$metric['keyDiff'] < 0)
+                        <small class="text-danger">{{ $metric['keyDiff'] }} % <i class="icon-arrow-down"></i></small>
+                    @elseif((float)$metric['keyDiff'] == 0)
+                        <small class="text-muted">{{ $metric['keyDiff'] }} % <i class="icon-refresh"></i></small>
```
tabuna: Not sure if we need an icon to indicate zero.  I copied this layout from the Yandex.Metrica service and there was zero as positive.  I have nothing against it

resources/views/container/layouts/metric.blade.php
```diff
@@ -8,12 +8,16 @@
...
                 <p class="h4 mb-1 text-black font-thin">{{ $metric['keyValue'] }}</p>
 
-                @if((float)$metric['keyDiff'] < 0)
-                    <small class="text-danger">{{ $metric['keyDiff'] }} % <i class="icon-arrow-down"></i></small>
-                @else
-                    <small class="text-success">{{ $metric['keyDiff'] }} % <i class="icon-arrow-up"></i></small>
+                @if(isset($metric['keyDiff']))
+                    @if((float)$metric['keyDiff'] < 0)
+                        <small class="text-danger">{{ $metric['keyDiff'] }} % <i class="icon-arrow-down"></i></small>
+                    @elseif((float)$metric['keyDiff'] == 0)
```
tabuna: I think that in the future it will be possible to change this place to use the operator of the ship.

resources/views/container/layouts/metric.blade.php
Nks: Yes, forgot about this ability. Agree.

resources/views/container/layouts/metric.blade.php
Nks: I think this is good to indicate that the data not changed.

tabuna: Probably you did not understand me. I meant exactly the `refresh icon`

PR #868, opened 2019-05-20
Change chart on chartjs

PSalador:


mojtabaahn:
Hey, I'm just curious is there any benefits in switching libraries?

tabuna:
> Hey, I'm just curious is there any benefits in switching libraries?

No, this is an alternate aerodrome, since for a long time the library we used was without support. 
Its OK now.

Issue #870, opened 2019-05-28
TD filter()  from related model


brahimbim:
Hello, how I can filter from related model 
![image](https://user-images.githubusercontent.com/24649160/58477063-26c85a80-8142-11e9-9b3d-7f37764d5e7a.png)

### Specifications

  - ORCHID Version: 4.7
  - Laravel Version: 5.8
  - PHP Version: 7.2
  - Database Version: 5.7

tabuna:
Hi @brahimbim 

This is not the most appropriate solution, we do not allow the use of HTTP filters for related models.

The best option would be to use a regular filter, where you can already set a modifying query. 

We have no plans to allow filtering, sorting for related models, but we can accept requests to add it.

PR #871, opened 2019-05-28
Change in httpFilter

brahimbim:
Fixes #

My proposition related to issue [#870](https://github.com/orchidsoftware/platform/issues/870)


tabuna:
@brahimbim Thank. I have a few questions:

- If I have some connection with the user, for example, comments -> users, will I be able to sort by his balance? And filtering by one-time password recovery code?
- I added a couple of tests in order to make sure that filtering and sorting works by json, could you check this by making a rebase master?

rajabilal555:
Why was this not Merged? Are there any issues?

PR #878, opened 2019-05-31
Added support of async modal for the button field

Nks:
I think we need loadModalAsync for the button because sometimes you need to send data to the modal but if we using async modal it overriding attributes which should be sent with the params.

### DIFF
--- a/resources/views/fields/button.blade.php
+++ b/resources/views/fields/button.blade.php
@@ -3,6 +3,8 @@
         <button type="button"
                 class="{{ $class }}"
                 data-action="screen--base#targetModal"
+                data-async="{{ ($async ? 'true' : 'false') }}"
+                @if($async)data-modal-params="@json($async_params)"@endif
                 data-modal-title="{{ $title ?? '' }}"
                 data-modal-key="{{ $modal ?? '' }}"
                 data-modal-action="{{ url()->current() }}/{{ $method }}"
--- a/resources/views/fields/button.blade.php
+++ b/resources/views/fields/button.blade.php
@@ -3,9 +3,9 @@
         <button type="button"
                 class="{{ $class }}"
                 data-action="screen--base#targetModal"
-                data-async="{{ ($async ? 'true' : 'false') }}"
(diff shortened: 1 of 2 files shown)


resources/views/fields/button.blade.php
```diff
@@ -3,6 +3,8 @@
         <button type="button"
                 class="{{ $class }}"
                 data-action="screen--base#targetModal"
+                data-async="{{ ($async ? 'true' : 'false') }}"
```
tabuna: Strict types allow you to make functions like this:

```
data-async="{{ var_export($async) }}"
```

resources/views/fields/button.blade.php
```diff
@@ -3,6 +3,8 @@
...
+                @if($async)data-modal-params="@json($async_params)"@endif
```
tabuna: Let's not write in style `snake case`

src/Screen/Fields/Button.php
```diff
@@ -124,9 +126,20 @@ public function type(string $visual): self
...
+        ], '', (string)$this->get('class'));
+
+        $this->set('class', $class . ' ' . $visual);
+
+        return $this;
+    }
 
-        $this->set('class', $class.' '.$visual);
+
+    public function loadModalAsync(string $modal, $method, $options, string $text = null): self
```
tabuna: Let's try to keep a single name in the executable file and the template file.
Also the absence of phpdocs makes the whole method look.


Let's write it more strictly, for example, options are declared by default as an array, 
why don't we allow them as `array $options = []`

resources/views/fields/button.blade.php
Nks: Fixed

src/Screen/Fields/Button.php
Nks: Added phpdoc and wrapper

resources/views/fields/button.blade.php
Nks: Rewrite in camelCase

PR #905, opened 2019-07-07
fix: problem in index

brahimbim:
array_merge change value of indexes in empty methode

### DIFF
--- a/src/Screen/Fields/Select.php
+++ b/src/Screen/Fields/Select.php
@@ -152,10 +152,7 @@ class Select extends Field
                 $options = $options->toArray();
             }
 
-            $value = array_merge(
-                [$key => $name],
-                $options
-            );
+            $value = [$key => $name] + $options;
 
             $this->set('options', $value);
         });


tabuna:
Hi @brahimbim 

Thank you for your interest in the development and corrections. But could you give me an example of a problem?

```php
$options = [1 => 'myValue'];

$key = 0;
$name = 'Default';

array_merge(
    [$key => $name],
    $options
);

/*
    array:2 [▼
      0 => "Default"
      1 => "myValue"
    ]
 */


[$key => $name] + $options;

/*
    array:2 [▼
      0 => "Default"
      1 => "myValue"
    ]
 */
```

brahimbim:
Hi @tabuna,

image
i use empty in this select field

image
i get false keys

tabuna:
I was able to confirm the problem and accepted the PR.
I also wrote a test so that in the future there would be no such problem.


https://github.com/orchidsoftware/platform/blob/0ec98592a2e70ec5055fef8f32666bccab71c740/tests/Unit/Fields/SelectTest.php#L156-L171

PR #908, opened 2019-07-10
Upload field review

Nks:
- Fixed issue with the multiple files
- Fixed issue with the multiple fields on one page
- Fixed issue with the maxFiles
- Added alerts on max files reached
- Fixed reorder issue
- Added terser options for the webpack. Now console.log will be dropped on the live.


resources/js/controllers/fields/upload_controller.js
```diff
@@ -236,15 +248,41 @@ export default class extends Controller {
...
+                });
+
+                this.on('sending', (file, xhr, formData) => {
+                    console.log('dropzone.sending');
+                    formData.append('_token', $('meta[name=\'csrf_token\']').attr('content'));
+                    formData.append('storage', storage);
+                    formData.append('group', groups);
+                });
+
+                this.on('removedfile', file => {
```
tabuna: What do you think of such a record?
```js
this.on('removedfile', file => {
  console.log('dropzone.removedfile');
  if (!objHas(file, 'data.id')) {
    return;
  }

  $(`${dropname} .files-${file.data.id}`).remove();

  !isMediaLibrary && axios
    .delete(platform.prefix(`/systems/files/${file.data.id}`), {
      storage: $('#post-attachment-dropzone').data('storage'),
    })
    .then();
});
```

tabuna:
Do we really need logging for development devs?
If there is a big screen with multiple fields, will this be a good solution?

I find out because I have not seen such an application.

resources/js/controllers/fields/upload_controller.js
```diff
@@ -236,15 +248,41 @@ export default class extends Controller {
...
+                    console.log('dropzone.removedfile');
+                    if (objHas(file, 'data.id')) {
+                        $(`${dropname} .files-${file.data.id}`).remove();
+                        if (!isMediaLibrary) {
+                            axios
+                                .delete(platform.prefix(`/systems/files/${file.data.id}`), {
+                                    storage: $('#post-attachment-dropzone').data('storage'),
```
tabuna: We can not specify a previously created variable?
```js
 storage:storage
 //or
 storage 
```

resources/js/controllers/fields/upload_controller.js
Nks: Оно, конечно, будет работать, но потом читабельность на нуле.
Это как `$boolean = $var1 ? $var1 : ($var2 ? $var2 : ($var3 ? $var3 : false);` 
Вроде бы работает, но с первого раза понять что происходит очень сложно. К тому же не добавляет никакого профита - из-за пары лишних строк, которые будут ужаты терсером в последующем никакого прироста производительности не будет.

tabuna: Я привел как пример, `!isMediaLibrary &&` давай тогда будем придерживаться `if` в этом случаи, хорошо?

Nks:
> Do we really need logging for development devs?
> If there is a big screen with multiple fields, will this be a good solution?
> We could connect Laravel Dusk to constantly check such things.
> 
> I find out because I have not seen such an application.

Technically - no. But I'm going to rewrite this field fully. It looks really worth for now and it better to have logs right now which will be cleared by the terser, but for development purposes while the field 
not yet completed fully I wonder to keep the logs. Also currently upload field doesn't work correctly.

resources/js/controllers/fields/upload_controller.js
Nks: Yep, we can use it too. But I am not paying attention for the quality for now cuz this fix just fixing the main logic. I am working on rewrute the upload field fully.

Nks:
updates? :)

PR #909, opened 2019-07-11
add localization to notification screen links

mojtabaahn:


src/Platform/Http/Screens/NotificationScreen.php
```diff
@@ -55,11 +55,11 @@ public function query(Request $request): array
     public function commandBar(): array
     {
         return [
-            Link::name('Remove all')
+            Link::name(ــ('Remove all'))
```
tabuna: Problem

Issue #915, opened 2019-07-24
Exposing js/sass files for customization


mojtabaahn:
Hi
I'm wondering if you have plans for exposing js/sass files into resources folder **optionally**, using `--tag=assets`.
after publishing this assets, need to add platform's package.json requirements to root package.json using a single `npm install ...` and then adding webpack.mix.js configurations.

This can be useful for per-project customization of front-end behaviors. 
Currently it needs to be copied and configured by user completely.

tabuna:
Yes it is in the plans

mojtabaahn:
If it is the same routine as I described I can make a PR.

tabuna:
I think we can do the same as laravel does with its presets. For example https://github.com/laravel/ui/blob/master/src/Presets/Bootstrap.php

I will be glad if you do this.

tabuna:
Hi, I returned to this topic again and am ready to work it out more thoroughly.

I think you need to make 2 presets:
- To work as a stimulus.js user, without changing the source code
- And more advanced already with the publication of all source code on js and css


Accordingly, there will be two command `artisan preset orchid` and `artisan preset orchid-source`

mojtabaahn:
I was thinking it is some sort of misleading to call them `presets` since it is actually possible to use them in existing assets and they do not remove all existing assets.

tabuna:
Yes it is, but in reality when the user does: `php artisan vendor:publish` for `view/css/js/etc` he refuses to receive updates, because his files will be used

PR #916, opened 2019-07-24
Adds ability to publish Orchid's asset files

mojtabaahn:
[#915](https://github.com/orchidsoftware/platform/issues/915)

### DIFF
--- a/src/Platform/OrchidPreset.php
+++ b/src/Platform/OrchidPreset.php
@@ -1,0 +1,97 @@
+<?php
+
+
+namespace Orchid\Platform;
+
+
+use Illuminate\Foundation\Console\Presets\Preset;
+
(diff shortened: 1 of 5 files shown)


src/Platform/OrchidPreset.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Install the preset.
+     *
+     * @return void
+     */
+    public static function install()
+    {
+        static::updatePackages(true);
+        static::updatePackages(false);
```
tabuna:
```php
static::updatePackages(true);
static::updatePackages(false);
```

so it should be?

mojtabaahn: It applies both devDependencies and dependencies in their respected steps.

PR #921, opened 2019-07-30
Adds stimulusjs preset (artisan preset orchid)

mojtabaahn:
#915

### DIFF
--- a/resources/views/app.blade.php
+++ b/resources/views/app.blade.php
@@ -21,9 +21,6 @@
     <script src="{{ orchid_mix('/js/manifest.js','orchid') }}" type="text/javascript"></script>
     <script src="{{ orchid_mix('/js/vendor.js','orchid') }}" type="text/javascript"></script>
     <script src="{{ orchid_mix('/js/orchid.js','orchid') }}" type="text/javascript"></script>
-    @if(file_exists(public_path('js/dashboard.js')))
-        <script src="{{ mix('/js/dashboard.js') }}" type="text/javascript"></script>
-    @endif
 
     @foreach(Dashboard::getResource('stylesheets') as $stylesheet)
         <link rel="stylesheet" href="{{  $stylesheet }}">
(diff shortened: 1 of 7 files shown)


resources/views/app.blade.php
```diff
@@ -21,6 +21,9 @@
     <script src="{{ orchid_mix('/js/manifest.js','orchid') }}" type="text/javascript"></script>
     <script src="{{ orchid_mix('/js/vendor.js','orchid') }}" type="text/javascript"></script>
     <script src="{{ orchid_mix('/js/orchid.js','orchid') }}" type="text/javascript"></script>
+    @if(file_exists(public_path('js/dashboard.js')))
+        <script src="{{ mix('/js/dashboard.js') }}" type="text/javascript"></script>
```
tabuna: We can use a more suitable extension method:

```php
namespace App\Providers;

use Illuminate\Support\ServiceProvider;
use Orchid\Platform\Dashboard;

class AppServiceProvider extends ServiceProvider
{
    /**
     * Bootstrap any application services.
     *
     * @return void
     */
    public function boot(Dashboard $dashboard)
    {
        $dashboard->registerResource('stylesheets', 'custom.css');
        $dashboard->registerResource('scripts', 'custom.js');
    }
}
```

Or adding to config

```php
'resource' => [
    'stylesheets' => [],
    'scripts'     => [],
],
```

Or, in extreme cases, the user can simply place it in his blade header template, override it to his:

```php
'template' => [
    'header' => 'custom.header',
    'footer' => 'platform::footer',
],
```

```php
//  view/custom/header.blade.php
{{ mix('/js/dashboard.js') }}
```

There are really a lot of options, do we need another solution?

tabuna: If you can’t use any of these solutions, then I suggest just asking the user about it as we do with npm.

mojtabaahn: Actually I was struggling how to achieve cleaner approach.

Last option seems very good IMO, What do you think
```
Please run "npm install && npm run dev" to compile your fresh scaffolding.
You need to also add below line to AppServiceProvider's register method after that:
app(\Orchid\Platform\Dashboard::class)->registerResource('scripts','/js/dashboard.js'); 
```

tabuna: I think that everything is more than good.


People will take it well, and they will do it with awareness. And even if not, then nothing terrible will happen and their project will not be broken.

Issue #927, opened 2019-08-08
TextArea not save previously entered value when Redirect::back()->withInput() used


likemusic:
### Expected Behavior
TextArea contains previously entered value.

### Actual Behavior
TextArea is empty for new entity and contains real value for exists entitities.

### Steps to Reproduce the Problem

  1. Use `TextArea::make('')` in layout() return value for screen (inside Layout::rows([])).
  1. In public function createOrUpdate(Product $product, Request $request)  use like:
```
try {
//do some things
throw new \Exception();
} catch (\Exception $e) {
            Alert::error($exception->getMessage());
            return Redirect::back()->withInput();
}
```

### Specifications

  - ORCHID Version: 5.5.0
  - Laravel Version: 5.8.30
  - PHP Version: 7.2.21
  - Database Version: mariadb:10.4

tabuna:
Hi @likemusic . Thanks for paying attention to this.

I prepared a change, you can check if it works for your scenario

tabuna:
It is believed that the problem is resolved.
If you run into this, feel free to reopen issue.

PR #928, opened 2019-08-08
Update Field.php

likemusic:
Possible fix for #927

### DIFF
--- a/src/Screen/Field.php
+++ b/src/Screen/Field.php
@@ -207,6 +207,7 @@ class Field implements FieldContract
             'slug'       => $this->getSlug(),
             'oldName'    => $this->getOldName(),
             'typeForm'   => $this->typeForm ?? $this->vertical()->typeForm,
+            'value'      => $this->modifyValue($this->getAttributes()['value']),
         ]))
             ->withErrors(session()->get('errors', app(ViewErrorBag::class)));
     }


tabuna:
I don’t think that creating additional variables is a good solution. We just deleted the part. Let's use an existing attribute in the view

likemusic:
@tabuna Existing attribute `value` for TextArea at the moment can't have old entered value bacause in getModifyAttributes() modifiers applyed only for universal attributes and inline attributes. TextArea have't `value` attribute in universal and inline attributes.

```
    /**
     * @return Collection
     */
    public function getModifyAttributes(): Collection
    {
        $modifiers = get_class_methods($this);

        collect($this->getAttributes())
            ->only(array_merge($this->universalAttributes, $this->inlineAttributes))
            ->each(function ($item, $key) use ($modifiers) {
                $signature = 'modify'.Str::title($key);

                if (in_array($signature, $modifiers, true)) {
                    $this->set($key, $this->$signature($item));
                }
            });

        return collect($this->getAttributes())
            ->only(array_merge($this->universalAttributes, $this->inlineAttributes));
    }
```

Should we delete collection->only() call at all, or add new field like $this-modifiedAttributes and merge it with $this->universalAttributes and $this->inlineAttributes?

tabuna:
Functionality of `modify*` attributes is now deprecated (from release 5.2.1 - 2019-07-11) and removed from the documentation. It was replaced by the opportunity to declare `addBeforeRender()`.
I think we can rework these 2 properties. 
By explicitly forming the name and value deleting everything else.

_Approximate option:_
```php
    public function render()
    {
        if (! $this->isSee()) {
            return;
        }

        $this->runBeforeRender();
        $this->checkRequired();
        $this->translate();
        $this->checkError();

        $id = $this->getId();
        $this->set('id', $id);

        $name = $this->getName();
        $this->set('name', $name);

        $value = $this->getValue();
        $this->set('value', $value);

        return view($this->view, [
            'attributes' => $this->getAttributes(),
            'id'         => $id,
            'old'        => $this->getOldValue(),
            'slug'       => $this->getSlug(),
            'oldName'    => $this->getOldName(),
            'typeForm'   => $this->typeForm ?? $this->vertical()->typeForm,
        ])
            ->withErrors(session()->get('errors', app(ViewErrorBag::class)));
    }
```

PR #951, opened 2019-09-11
Add force option for orchid:link

AlexSabur:


src/Platform/Commands/LinkCommand.php
```diff
@@ -32,20 +32,23 @@ class LinkCommand extends Command
      */
     public function handle(Dashboard $dashboard)
     {
-        $prefix = public_path('/resources');
+        $prefix = public_path('resources');
 
-        if (file_exists($prefix)) {
+        if (file_exists($prefix) && !$this->option('force')) {
```
tabuna: Hi, I think that `force` should also update the previously created links. 
I suggest simply if you have the option to delete the directory?

```php
if($this->option('force')) {
    $this->getLaravel()->make('files')->deleteDirectory($prefix)
}
```
Then there will be minimum changes and the update will be in any case.

What do you think about it?

tests/Console/ArtisanTest.php
```diff
@@ -96,6 +96,9 @@ public function testArtisanOrchidLink()
 
         $this->artisan('orchid:link')
             ->expectsOutput("The [$resources] directory already exists.");
+
+        $this->artisan('orchid:link --force')
```
tabuna: Let's create a separate test that created a directory and first received an error, and fixing it with --force

src/Platform/Commands/LinkCommand.php
AlexSabur: When deleting via `deleteDirectory`, there is a chance that it will delete the contents of the link (If the source is not from vendor)

tabuna: Do you mean it will not delete the link, but the content that was generated by the user? 
For example, commands like: `artisan preset orchid-source` ?

AlexSabur: For some reason, if the source of the link is not in the `vendor`, for example, `./packages/platform/public`, then when the `deleteDirectory` is called, the contents of `./packages/platform/public` will also be deleted. (Maybe this is a windows problem)

tabuna: Oh, good. Then let's leave it as it is.
But there is a point that `force` assumes that the link will be replaced, but in reality only new ones are added, while the old ones are not deleted. Can we do something about this?

PR #954, opened 2019-09-17
Make table responsive

osbre:
Make table responsive

Fixes mobile compatibility

## Proposed Changes

  - Support mobile compatibility
  - Inside docs: https://getbootstrap.com/docs/4.1/content/tables/#responsive-tables

### DIFF
--- a/resources/views/layouts/table.blade.php
+++ b/resources/views/layouts/table.blade.php
@@ -1,4 +1,4 @@
-<div class="row">
+<div class="row table-responsive-lg">
     <table class="table">
         <thead>
         <tr>


tabuna:
Hi, thanks for taking care of this, tell me if the column filters in this display work?

image

In particular, I worry about the calendar on a very short table

osbre:
@tabuna, in mobile, it works only if we have some data:

Знімок екрана  о 14 58 56

Знімок екрана  о 14 58 26


But in desktop, all okay:

Знімок екрана  о 14 59 26

osbre:
@tabuna, I was updated commit history, and now fixed this bug.

See screenshots:

Знімок екрана  о 17 23 29

PR #956, opened 2019-09-17
Support to customize app template

osbre:
## Proposed Changes

  - Allow to change app template, inside config
  - Allow to change container class, inside app template

## Usage example

For example, i can change my config to:

    'template' => [
        'header' => 'platform::header',
        'footer' => 'platform::footer',
        'layout' => 'app',
    ],

And create view named **app.blade.php** :

```
@extends('platform::app')

@section('container-class', 'container-fluid')
```

Then in this view, we can fully customize anything<del>, or just add a class to container :D </del>

### DIFF
--- a/config/platform.php
+++ b/config/platform.php
@@ -98,6 +98,7 @@ return [
     'template' => [
         'header' => 'platform::header',
         'footer' => 'platform::footer',
+        'layout' => 'platform::app',
     ],
 
 ];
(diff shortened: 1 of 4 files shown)


osbre:
Please, see https://github.com/orchidsoftware/platform/pull/944, add do not make similar changes, becouse we can just : 

php artisan vendor:publish --view

And change ANYTHINHG.

PR #957, opened 2019-09-18
Update modal.blade.php

2kodes:
This ensures params or options given to ModalToggle::loadModalAsync() are available in the request used to load the async modal.

Fixes #
Issue with options/params given to ModalToggle::loadModalAsync() not showing in the request


tabuna:
@2kodes Yes, let's change

PR #961, opened 2019-09-22
Fix scope not working in Relation

yxuko:
This fixes Relation scope not working

**Code**
`Relation::make('chief.')->fromModel(User::class, 'name')->applyScope('bureau')`

**Expected behavior**
Getting results and can search results

**Actual behavior**
Getting no results and doing a search results in :
`Cannot use object of type Illuminate\Database\Query\Builder as array`

### DIFF
--- a/src/Platform/Http/Controllers/Systems/RelationController.php
+++ b/src/Platform/Http/Controllers/Systems/RelationController.php
@@ -64,7 +64,7 @@ class RelationController extends Controller
             $model = $model->handler();
         }
 
-        $items = collect($model);
+        $items = collect($model->get());
 
         if (! is_null($search) && $search !== '') {
             $items = $items->filter(function ($item) use ($name, $search) {
--- a/src/Platform/Http/Controllers/Systems/RelationController.php
+++ b/src/Platform/Http/Controllers/Systems/RelationController.php
@@ -43,7 +43,7 @@ class RelationController extends Controller
     }
 
     /**
-     * @param object|Model $model
+     * @param array|object|Model $model
      * @param string       $name
      * @param string       $key
      * @param string       $search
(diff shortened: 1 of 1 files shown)


tabuna:
Hi, this is a good change, but it can cause problems in existing applications.

Where before it is written:

```php
public function scopeBureau() {
   return ...->get();
}
```

It can be called twice. What will lead to an error.
Can we try to add verification to satisfy both conditions?

yxuko:
Adding `->get()` in function scopeBureau() causes scope to return collection instead of query which is needed when calling `User::bureau()` in list screen

tabuna:
Ok, let's do the following. I wrote a test in a [branch](https://github.com/orchidsoftware/platform/tree/issue/961_relation)  that would check all possible return values:

```php
/**
 * @param \Illuminate\Database\Eloquent\Builder $query
 *
 * @return \Illuminate\Database\Eloquent\Builder
 */
public function scopeAsBuilder(Builder $query) : Builder
{
    return $query->where('name', 'RelationTest');
}

/**
 * @param \Illuminate\Database\Eloquent\Builder $query
 *
 * @return \Illuminate\Support\Collection
 */
public function scopeAsCollection(Builder $query) : Collection
{
    return $query->where('name', 'RelationTest')->get();
}

/**
 * @param \Illuminate\Database\Eloquent\Builder $query
 *
 * @return array
 */
public function scopeAsArray(Builder $query): array
{
    return $query->where('name', 'RelationTest')->get()->all();
}
```

https://github.com/orchidsoftware/platform/blob/d92df5ade70ce5667c415bfa6c0b5d7d2680fda7/tests/Feature/Platform/RelationsTest.php

Can you change this so that it works perfectly in all cases?

tabuna:
Thank!

PR #963, opened 2019-09-23
Update modal.blade.php

2kodes:
Fixes #
Issue where the submit button in the modal does not show a spinning animation when clicked and the form is being submitted

## Proposed Changes

Specify correct key for element to animate when form is being submitted or form submit button is clicked

Change loop key so it does not interfere with the key variable for the submit button key (without this the id of the submit button is `submit-modal-0`  instead of `submit-button-modalName`)

### DIFF
--- a/resources/views/layouts/modal.blade.php
+++ b/resources/views/layouts/modal.blade.php
@@ -16,7 +16,7 @@
                       enctype="multipart/form-data"
                       data-controller="layouts--form"
                       data-action="layouts--form#submit"
-                      data-layouts--form-button-animate="#screen-modal-{{$key}}"
+                      data-layouts--form-button-animate="#submit-modal-{{$key}}"
                       data-layouts--form-button-text="{{ __('Loading...') }}"
                 >
                     <div class="modal-header">
@@ -26,7 +26,7 @@
                     </div>
                     <div class="modal-body">
                         <div data-async>
-                            @foreach($manyForms as $key => $modal)
+                            @foreach($manyForms as $formKey => $modal)
                                 @foreach($modal as $item)
                                     {!! $item ?? '' !!}
                                 @endforeach


resources/views/layouts/modal.blade.php
```diff
@@ -16,7 +16,7 @@
                       enctype="multipart/form-data"
                       data-controller="layouts--form"
                       data-action="layouts--form#submit"
-                      data-layouts--form-button-animate="#screen-modal-{{$key}}"
+                      data-layouts--form-button-animate="#submit-modal-{{$key}}"
```
tabuna: We are sending another form:
```php
  <form class="modal-content"
    id="screen-modal-form-{{$key}}"
    method="post"
```
What is this change for?

2kodes: for animating the submit button when clicked..

when a submit button is clicked, in the submit action/method of layouts/form_controller.js you call the function ` this.animateButton();` 

`animateButton() {
        const button = this.data.get('button-animate');
        const text = this.data.get('button-text') || '';`

tabuna: I understand the animation. I want to say that the forms with the name `#submit-modal-{{$key}}`  does not exist

2kodes: oh ok.
 So here this `data-layouts--form-button-animate="#submit-modal-{{$key}}"` is not referring to a form but rather storing the attribute or value `button-animate` in the stimulusjs  controller. 

Please [check here](https://stimulusjs.org/reference/data-maps.html) for reference

You are getting those attributes or controller data in this line from the file `form_controller.js`

`const button = this.data.get('button-animate');
 const text = this.data.get('button-text') || '';`

tabuna: Yes, that’s right, it’s my fault that I tried to mislead you for myself.

PR #1026, opened 2019-12-09
Strict comparison operator

osbre:
Fixes this error:

`Object of class Illuminate\Database\Eloquent\Collection could not be converted to int (View: /Users/ostap/Desktop/smart-logistics/resources/views/vendor/platform/fields/select.blade.php)`

## Proposed Changes

  - Change comparison to strict comparison

## Steps to reproduce

Create some screen, and add this field:

```php
Select::make('delivery.users.')
    ->title(__('Managers'))
    ->options(
        User::all()->pluck('name', 'id')
    )->multiple()
```

### DIFF
--- a/resources/views/fields/select.blade.php
+++ b/resources/views/fields/select.blade.php
@@ -6,7 +6,7 @@
                         @isset($value)
                         @if (is_array($value) && in_array($key, $value)) selected
                         @elseif (isset($value[$key]) && $value[$key] == $option) selected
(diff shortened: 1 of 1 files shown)


osbre:
Solved. Use instead:

```php
Select::make('delivery.users.')
    ->title(__('Managers'))
    ->multiple()
    ->fromModel(User::class, 'name')
    ->applyScope('manager')
```

PR #1033, opened 2019-12-17
Add the ability to change default migrations

osbre:
## Proposed Changes

  - Add more freedom to published migrations, for example - renaming.


osbre:
sorry, commits conflict

osbre:
I think this may damage some users' backwards compatibility.

I'll create another pull for 7.0 branch.

PR #1036, opened 2019-12-22
The name of the tables from the settings.

ivan-cc:

### DIFF
--- a/config/platform.php
+++ b/config/platform.php
@@ -110,7 +110,7 @@ return [
     */
 
     'rolesTable'            => env('ORCHID_ROLES_ROLES_DATABASE_TABLE', 'roles'),
-    'roleUserTable'         => env('ORCHID_ROLES_ROLE_USER_DATABASE_TABLE', 'role_users'),
+    'roleUserTable'         => env('ORCHID_ROLE_USERS_DATABASE_TABLE', 'role_users'),
     'settingsTable'         => env('ORCHID_SETTINGS_DATABASE_TABLE', 'settings'),
     'attachmentsTable'      => env('ORCHID_ATTACHMENTS_DATABASE_TABLE', 'attachments'),
     'attachmentableTable'   => env('ORCHID_ATTACHMENTABLE_DATABASE_TABLE', 'attachmentable'),
--- a/config/platform.php
+++ b/config/platform.php
@@ -99,5 +99,21 @@ return [
         'header' => 'platform::header',
         'footer' => 'platform::footer',
     ],
+    
+    /*
+    |--------------------------------------------------------------------------
+    | Package Connection
+    |--------------------------------------------------------------------------
+    |
+    | You can set your own names for tables.
+    |
+    */
+
+    'rolesTable'            => env('ORCHID_ROLES_ROLES_DATABASE_TABLE', 'roles'),
+    'roleUserTable'         => env('ORCHID_ROLES_ROLE_USER_DATABASE_TABLE', 'role_users'),
+    'settingsTable'         => env('ORCHID_SETTINGS_DATABASE_TABLE', 'settings'),
+    'attachmentsTable'      => env('ORCHID_ATTACHMENTS_DATABASE_TABLE', 'attachments'),
+    'attachmentableTable'   => env('ORCHID_ATTACHMENTABLE_DATABASE_TABLE', 'attachmentable'),
+    'notificationsTable'    => env('ORCHID_NOTIFICATIONS_DATABASE_TABLE', 'notifications'),    
 
 ];
--- a/database/migrations/2015_10_19_214424_create_orchid_roles_table.php
+++ b/database/migrations/2015_10_19_214424_create_orchid_roles_table.php
@@ -11,7 +11,7 @@ class CreateOrchidRolesTable extends Migration
      */
     public function up()
     {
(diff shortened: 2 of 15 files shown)


osbre:
When you install orchid, migrations automatically published, and can you customize anything (except file names, see #1034)

ivan-cc:
> Когда вы устанавливаете orchid, миграции автоматически публикуются, и вы можете настроить что угодно (кроме имен файлов, см. [# 1034](https://github.com/orchidsoftware/platform/pull/1034) )

Then the relationship will not work.

Example:
https://github.com/orchidsoftware/platform/pull/1036/files#diff-5b74d47d7a1195b2c74737b361a26ade

tabuna:
Hey @ivan-cc . Thank you for this work.

> Then the relationship will not work.

This is not true. In all respects, the class is not shown explicitly. And it can be redefined. 
This is stated in the [documentation](https://orchid.software/en/docs/configuration/#model-classes), here are some excerpts:


## Model classes

The desire to change the behavior of some classes from the standard delivery is quite normal, for the platform to use your model classes instead of its own, it is necessary to register their substitution in advance using:

```php
Dashboard::useModel(\Orchid\Platform\Models\User::class, \App\User::class);
```

You can use the configuration parameter, which allows you to define all the substitutions at once:

```php
Dashboard::configure([
    'models' => [
        User::class => MyCustomClass::class,
    ],
]);
```

------

Thus, you can replace classes from the package with your own if necessary.

ivan-cc:
> Hey @ivan-cc . Thank you for this work.
> 
> > Then the relationship will not work.
> 
> This is not true. In all respects, the class is not shown explicitly. And it can be redefined.
> This is stated in the [documentation](https://orchid.software/en/docs/configuration/#model-classes), here are some excerpts:
> 
> ## Model classes
> The desire to change the behavior of some classes from the standard delivery is quite normal, for the platform to use your model classes instead of its own, it is necessary to register their substitution in advance using:
> 
> ```
> Dashboard::useModel(\Orchid\Platform\Models\User::class, \App\User::class);
> ```
> 
> You can use the configuration parameter, which allows you to define all the substitutions at once:
> 
> ```
> Dashboard::configure([
>     'models' => [
>         User::class => MyCustomClass::class,
>     ],
> ]);
> ```
> 
> Thus, you can replace classes from the package with your own if necessary.

Override class for one line of code?
Not a good solution.

What is bad to use settings?

tabuna:
> Override class for one line of code?
> Not a good solution.

This is customary in the Laravel ecosystem. For example, you can see the spark [documentation](https://spark.laravel.com/docs/9.0/customization#model-classes)

> What is bad to use settings?

This is not a bad solution. But we already have the means to redefine the classes inside the package.
What are the additional settings for? What do they give the user?

Also, the current version will break backward compatibility, since already installed settings do not have these parameters, and the default value is not specified.

Let's discuss it

ivan-cc:
For example, like this:
`config('platform.rolesTable', 'roles');`

It will be possible to maintain compatibility.

ivan-cc:
This will give the user the ability to easily resolve conflicts with other laravel packages.

For example, I use the [laravel-auth](https://github.com/jeremykenedy/laravel-auth) package, which has table `roles`.

There are many more packages that use `attachments` or `settings` names for tables.

The Orchid package will not be installed due to a conflict in the tables.

tabuna:
@ivan-cc It’s still not clear to me how packages with the same functionality will avoid conflicts without a large code base on both sides.

Using the role model as an example, I’ll try to explain how this can be done in my own application.
Install a clean Laravel template and Orchid package. Let's migrate and create a new one:

```php
class RoleRenameTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::rename('roles', 'orchid_roles');
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::rename('orchid_roles', 'roles');
    }
}

```

Add your own role class:

```php
namespace App;

class Role extends \Orchid\Platform\Models\Role
{
    /**
     * The database table used by the model.
     *
     * @var string
     */
    protected $table = 'orchid_roles';
}
```

Making a substitution for vendor:

```php
namespace App\Providers;

use Illuminate\Support\ServiceProvider;
use Orchid\Support\Facades\Dashboard;

class AppServiceProvider extends ServiceProvider
{
    /**
     * Bootstrap any application services.
     *
     * @return void
     */
    public function boot()
    {
        Dashboard::useModel(\Orchid\Platform\Models\Role::class, \App\Role::class);
    }
}
```

It remains only to change in the presented installation files to your model.

infostreams:
I have an already existing 'roles' table, because of the very popular Spatie package "laravel-permission". Adding a migration to rename that existing table, then run the Orchid installer, then add another migration to rename the Orchid 'roles' table... that's insane.

Perhaps you could take inspiration from https://github.com/spatie/laravel-permission/blob/master/database/migrations/create_permission_tables.php.stub to see how they do it? I would personally prefer if you could just namespace ('prefix') all of your tables by default, so that they would be named 'orchid_roles'. Or at the very least make them configurable.

osbre:
In orchid 7, you will be able to publish migrations and change even file names, becouse they not load from package

tabuna:
The settings table is no longer there.

PR #1059, opened 2020-03-05
Not required icon element

osbre:
Icon require a bit space if not set


osbre:
Closed through https://github.com/orchidsoftware/platform/pull/1063

PR #1074, opened 2020-04-18
add default to user permissions

G3z:
Fixes #
When trying to access the dashboard with an empty permissions field the error below is received, while setting a default ([] or {}) results in a 403 which is the expected behaviour
`Argument 1 passed to Orchid\Platform\Models\User::Orchid\Access\{closure}() must be of the type array, null given`
## Proposed Changes
  - add a default to the column
  - check if `$this->cachePermissions` is null in `src/Access/UserAccess.php:77`

### DIFF
--- a/database/migrations/2015_04_12_000000_create_orchid_users_table.php
+++ b/database/migrations/2015_04_12_000000_create_orchid_users_table.php
@@ -13,7 +13,7 @@ class CreateOrchidUsersTable extends Migration
     {
         Schema::table('users', function (Blueprint $table) {
             $table->timestamp('last_login')->nullable();
-            $table->jsonb('permissions')->default('{}')->nullable();
+            $table->jsonb('permissions')->nullable();
         });
     }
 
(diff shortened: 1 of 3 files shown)


tabuna:
Hi @G3z , thanks for paying attention to this.
I think you should not change the migration because current users will not receive this fix. 
We can change this as you said in the verification files:

```php
// src/Access/UserAccess.php:69

if (! $cache || is_null($this->cachePermissions)) {
    $this->cachePermissions = $this->roles()
        ->pluck('permissions')
        ->prepend($this->permissions)
        ->filter(function ($permission){
            return is_array($permission);
        });
}
```
And a rough test:
```php
// platform/tests/Unit/PermissionTest.php

public function testEmptyPermission(): void
{
    $nullPermission = $this->createUser()
        ->setAttribute('permission', null)
        ->hasAccess('access.to.secret.data');
        
    $stringPermission = $this->createUser()
        ->setAttribute('permission', '')
        ->hasAccess('access.to.secret.data');

    $this->assertFalse($nullPermission);
    $this->assertFalse($stringPermission);
}
```

What do you say about this decision?

G3z:
Hello @tabuna, I moved the code as you suggested, thank you!

tabuna:
The release is ready: [7.0.12](https://github.com/orchidsoftware/platform/releases/tag/7.0.12)

G3z:
thank you

PR #1075, opened 2020-04-19
custom model attributes in selects

G3z:
by executing the query before plucking model attributes can be used as 'name' property

Fixes #
```php
Select::make('category.parent_id')
                ->title('Categoria padre')
                ->fromQuery(Category::byCurrentShop()->orderBy('path'), 'indented_name') // indented_name is magic attribute of Category (getIndentedNameAttribute),
```
## Proposed Changes
- Execute the query and perform the pluck on models

### DIFF
--- a/src/Screen/Fields/Select.php
+++ b/src/Screen/Fields/Select.php
@@ -133,7 +133,7 @@ class Select extends Field
     {
         $key = $key ?? $builder->getModel()->getKeyName();
 
-        return $this->setFromEloquent($builder, $name, $key);
+        return $this->setFromEloquent($builder->get(), $name, $key);
     }
 
     /**
--- a/tests/Unit/Screen/Fields/SelectTest.php
+++ b/tests/Unit/Screen/Fields/SelectTest.php
@@ -1,12 +1,13 @@
 <?php
 
-declare(strict_types=1);
+declare (strict_types = 1);
 
 namespace Orchid\Tests\Unit\Screen\Fields;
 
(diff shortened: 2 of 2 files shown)


tabuna:
Hi @G3z . Thanks.
 Can you add a test to be sure that it will not be broken in the future?

G3z:
I did this, but how can I test that `'name'` can be a runtime attribute ?

```php
public function testAttributesCanBeUsed()
    {

        $select = Select::make('choice')
            ->empty('empty')
            ->fromQuery(Role::orderBy('id'), 'name');

        $view = self::minifyRenderField($select);

        foreach ($this->roles as $role) {
            $option = $this->stringOption($role->name, $role->id);
            $this->assertStringContainsString($option, $view);
        }

        $option = $this->stringOption('empty');
        $this->assertStringContainsString($option, $view);
    }
```

also, maybe a function can be cool
```php
$select = Select::make('choice')
            ->empty('empty')
            ->fromQuery(Role::orderBy('id'), function($row){
                  return [$row->key => $row->title];
            });
```

tabuna:
I think we can do the same thing as [checking the calculated](https://github.com/orchidsoftware/platform/blob/91699679277c1cdbce4c39bfc42ebf1432937158/tests/Unit/Screen/Fields/RelationTest.php#L1) field of the `Relation` field :

```php
$select = Select::make('choice')
            ->empty('empty')
            ->fromQuery(EmptyUserModel::orderBy('id'), 'full');
```

PR #1077, opened 2020-04-20
Auto added dot for multiple

tabuna:
This change makes the point at the end of the field name optional for the multiple.

_Before_
```php
Select::make('user.roles.')
    ->fromModel(Role::class, 'name')
    ->multiple()
```

_After_

```php
Select::make('user.roles')
    ->fromModel(Role::class, 'name')
    ->multiple()
```

For those places where there is already a specified dot, the work will be unchanged.

### DIFF
--- a/src/Screen/Concerns/Multipliable.php
+++ b/src/Screen/Concerns/Multipliable.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Orchid\Screen\Concerns;
+
+use Illuminate\Support\Str;
+
+trait Multipliable
+{
+    /**
+     * @return self
+     */
+    public function multiple(): self
+    {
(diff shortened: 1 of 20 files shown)


tabuna:
Unfortunately, this will probably do more harm than good. Because the binding is carried out in the builder and for the forms used in them there will be no problems. 

But if you form a line explicitly get:

```php
(string) Select::make('roles')
    ->fromModel(Role::class, 'name')
    ->multiple()

// <select name="roles." multiple>
```

In the same thing in the builder
```php
// <select name="roles[]" multiple>
```

Accordingly, normal use does not work:
```php
(string) Select::make('roles[]')
    ->fromModel(Role::class, 'name')
    ->multiple()

// <select name="roles[]." multiple>
```

PR #1084, opened 2020-04-27
refs #774 [Layouts] Init Listener

### LINKED ISSUE #774: [Draft] Conditional fields
This will be really helpful add an ability to set the conditional fields which can appear only when it really needed. I know that this is possible with the rules, but in the screens page still overloaded with a lot of not needed all time fields.

Open to discuss

tabuna:
Fixes #774


tabuna:
In order to create a new listening layer, you need to run the artisan command:
```bash
php artisan orchid:listener SumListener
```
As a result, a new class will be created in the `app/Orchid/Layouts/SumListener.php` directory.


After that, we can use it in our screen, for example:

```php
namespace App\Orchid\Screens;

use Orchid\Screen\Fields\Input;
use Orchid\Screen\Layout;
use Orchid\Screen\Screen;

class SumScreen extends Screen
{
    /**
     * Display header name.
     *
     * @var string
     */
    public $name = 'Sum';
    
    /**
     * Query data.
     *
     * @return array
     */
    public function query(): array
    {
        return [];
    }

    /**
     * Button commands.
     *
     * @return Action[]
     */
    {
        return [];
    }

    /**
     * Views.
     *
     * @return Layout[]
     */
    public function layout(): array
    {
        return [

            Layout::rows([
                Input::make('a')
                    ->title('First argument')
                    ->type('number'),

                Input::make('b')
                    ->title('Second argument')
                    ->type('number'),
            ]),
        ];
    }
}
```

We want to display the result of adding the two fields presented. To do this, specify their names:

```php
namespace App\Orchid\Layouts;

use Orchid\Screen\Layout;
use Orchid\Screen\Layouts\Listener;

class SumListener extends Listener
{
    /**
     * @var string[]
     */
    protected $targets = [
        'a',
        'b',
    ];

    /**
     * What screen method should be called
     * as a source for an asynchronous request.
     *
     * The name of the method must
     * begin with the prefix "async"
     *
     * @var string
     */
    protected $asyncMethod = '';

    /**
     * @return Layout[]
     */
    protected function layouts(): array
    {
        return [];
    }
}

```

But this is not enough, we must define a method on the screen that would add our values:

```php
class SumScreen extends Screen
{
    //...
    
    public function asyncSum($a, $b)
    {
        return [
            'sum' => $a + $b,
        ];
    }

    //...
}
```

We specify this method in the listener class:
```php
class SumListener extends Listener
{
    /**
     * What screen method should be called
     * as a source for an asynchronous request.
     *
     * The name of the method must
     * begin with the prefix "async"
     *
     * @var string
     */
    protected $asyncMethod = 'asyncSum';
    
    // ...
}
```

Now we listen to two fields and add their results. It remains to add input:

```php
class SumListener extends Listener
{
    /**
     * @return Layout[]
     */
    protected function layouts(): array
    {
        return [
            Layout::rows([
                Input::make('sum')
                    ->title('SUM')
                    ->help('The result of adding the first argument and the second')
                    ->canSee($this->query->has('a') && $this->query->get('b'))
            ]),
        ];
    }
}
```

Display a new layer on the screen:
```php
class SumScreen extends Screen
{
    //...
    
    /**
     * Views.
     *
     * @return Layout[]
     */
    public function layout(): array
    {
        return [
            Layout::rows([
                Input::make('a')
                    ->title('First argument')
                    ->type('number'),

                Input::make('b')
                    ->title('Second argument')
                    ->type('number'),
            ]),

            SumListener::class
        ];
    }
    
    //...
}
```

PR #1109, opened 2020-05-28
Add resizing options to Cropper

### LINKED ISSUE #1108: More flexibility with Cropper
**Is your feature request related to a problem? Please describe.**
At the moment, if I'm not mistaken, only with and height are settable on Cropper but i noticed Cropper.js has a few more options https://github.com/fengyuanchen/cropperjs#getcroppedcanvasoptions

**Describe the solution you'd like**
I believe it would be easy to implement `minWidth` `minHeight` `maxWidth` `maxHeight` (and also `max` and `min` shortcuts).
If this makes sense to you I will make a PR

G3z:
Fixes #1108

## Proposed Changes
added four methods to Cropper
```
 * @method Cropper minWidth($value = null)
 * @method Cropper minHeight($value = null)
 * @method Cropper maxWidth($value = null)
 * @method Cropper maxHeight($value = null)
 * @method Cropper min($value = null)
 * @method Cropper max($value = null)
```
I'm not really fond of `min` and `max` alternatives would be:
- `minSize`/`maxSize`
- `minShort`/`maxLarge`


tabuna:
Hi @G3z. It’s not very obvious to me what methods do `min` and `max`.
Adding `minSize` does not really help. Anyway, you can think about the actual size in `mb`

src/Screen/Fields/Cropper.php
```diff
@@ -30,6 +30,12 @@
  * @method Cropper popover(string $value = null)
  * @method Cropper title(string $value = null)
  * @method Cropper maxFileSize($value = true)
+ * @method Cropper minWidth($value = null)
+ * @method Cropper minHeight($value = null)
+ * @method Cropper maxWidth($value = null)
+ * @method Cropper maxHeight($value = null)
+ * @method Cropper min($value = null)
+ * @method Cropper max($value = null)
```
tabuna, line 38: These are real methods, there is no need to specify additional meta attributes

G3z: cleared!

G3z:
how about `minShort` / `maxLarge` ?

tabuna:
I think this is better. But still not sure, maybe a `minCanvas/maxCanvas`?

G3z:
fine by me

PR #1110, opened 2020-05-28
add storage to Cropper and Image metadata

G3z:
## Proposed Changes
Add `storage` metadata to Picture and Cropper


tabuna:
Shouldn't we pass this value to js and to the server?

G3z:
i did a quick test while tinkering with cropper and it worked already

tabuna:
I'm sorry.  it has already been done


https://github.com/orchidsoftware/platform/blob/2ab09f6c4a1e6c37b40892576aab4ed1e9b8138c/resources/js/controllers/fields/picture_controller.js#L53

PR #1113, opened 2020-05-29
print roles as default user subtitle

G3z:
I know this is a minor thing and users can easily customize this but still I think this is a cool default

## Proposed Changes
roles as default subtitle for userpresenter

### DIFF
--- a/install-stubs/Orchid/Presenters/UserPresenter.php
+++ b/install-stubs/Orchid/Presenters/UserPresenter.php
@@ -32,7 +32,7 @@ class UserPresenter extends Presenter implements Searchable, Personable
      */
     public function subTitle(): string
     {
-        return $this->entity->roles->pluck('name')->implode(' - ');
+        return $this->entity->roles->pluck('name')->implode(' / ');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tabuna:
I think between `Regular user` and `Plain user` at the request of Google, the first phrase is more common

install-stubs/Orchid/Presenters/UserPresenter.php
```diff
@@ -32,7 +32,11 @@ public function title(): string
      */
     public function subTitle(): string
     {
-        return 'Administrator';
```
tabuna, line 35: What do you think of this change?

install-stubs/Orchid/Presenters/UserPresenter.php
```diff
@@ -32,7 +32,11 @@ public function title(): string
...
+        $roles = $this->entity->roles->pluck('name')->implode(' / ');
```
tabuna, line 35: What do you think of this change?

G3z: Fine by me!

G3z:
i'll add the `?? 'Regular user'` fallback

tabuna:
Adding `?? 'Regular user'` does not work as an empty string will be returned

PR #1133, opened 2020-06-08
handle options from cropper

G3z:
Fixes https://github.com/orchidsoftware/platform/pull/1110

if only width and height are specified cropper currently does nothing. by passing only defined parameters cropper works as intended


tabuna:
Hi @G3z  Don't such default settings solve the problem?

https://github.com/orchidsoftware/platform/blob/1fd9f77c6cf9a27edcb5d6a4efc0e13817e7bbc4/src/Screen/Fields/Cropper.php#L47-L58

G3z:
I finally had time to check, Yes it does, closing the PR

PR #1135, opened 2020-06-08
Spanish translation modifications

ricardoaponte:
Fixes #

## Proposed Changes

  -
  -
  -


resources/lang/es.json
```diff
@@ -0,0 +1,167 @@
+{
+  "\/about": "\/о нас",
```
tabuna, line 2: Russian word??

ricardoaponte: Where?

ricardoaponte: I saw it. It is the first word. Did not see it before. I will fix it.

PR #1164, opened 2020-06-26
Add missing property to the Dashboard class - currentScreen

MunGell:
## Proposed Changes

  - Add `$currentScreen` private property in `Dashboard` class as it is used in `setCurrentScreen` and `setCurrentScreen`. Throws PHP Notice in console, breaks CI tests


MunGell:
Thats strange, maybe I've branched from a wrong place, let me check

MunGell:
Looks like there was a force push to the branch

tabuna:
My thousand apologies, initially said the wrong thread and rewrote it

MunGell:
Was fixed already

PR #1174, opened 2020-07-09
Switch off autocomplete on "email" signin field for security reasons

MunGell:
## Issue

The login screen allows caching of usernames (emails).

## Impact

Caching usernames in the browser is considered bad practice as it reveals valid usernames to other users of the computer (e.g. when used on hotel public computers etc.)

## Solution

HTML modified to add `autocomplete="off"` attribute on the `email` field on login screen.

### DIFF
--- a/resources/views/auth/signin.blade.php
+++ b/resources/views/auth/signin.blade.php
@@ -7,6 +7,7 @@
         ->required()
         ->tabindex(1)
         ->autofocus()
+        ->autocomplete('off')
         ->placeholder(__('Enter your email'))
     !!}
 </div>


tabuna:
Hey @MunGell .

Thank you for paying attention to this. In fact, setting this option may be ignored by browsers and extensions.

- https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off
- https://bugs.chromium.org/p/chromium/issues/detail?id=914451

Google has also spoken on this subject since about 2014. 
I found various ways around this, for example:

```html
<input type="search" autocomplete="off">
```

And other similar tricks. Do we need to worry about this?

MunGell:
Good point, `autocomplete=off` wouldn't prevent browser from auto-filling login details if user saved them in the browser's password manager. However, it will prevent browser from suggesting usernames that were inputted in that field before.

However, I also noticed that companies like Google and GitHub set `autocomplete` to `username` on their login pages, which allow for showing cached usernames in that field.

Closing this for the moment until further clarification will be found on the correct value of it.

PR #1200, opened 2020-08-06
Update phpDoc in Button class

marad3r:
phpStan (PHP Static Analysis Tool) throws an error  #Call to an undefined method Orchid\\Screen\\Actions\\Button\:\:asyncParameters\(\)#, because phpDoc in method make declares to return Button, but code declares creating static class


tabuna:
Thank you!

PR #1232, opened 2020-09-15
Update Field.php

vladitot:
Fixes #

## Proposed Changes

  - Orchid has checks for allowing "data-" attributes, but I cant set them, because orchid filters attributes by allowed list (universal attributes and inline attributes). So, this fix is allowing to use "data-test" attribute in components without extending them.
 - adding "data-test" attribute. It is great solution to specify component names, for acceptance testing.

### DIFF
--- a/src/Screen/Field.php
+++ b/src/Screen/Field.php
@@ -106,6 +106,7 @@ class Field implements Fieldable
         'title',
         'xml:lang',
         'autocomplete',
+        'data-test',
     ];
 
     /**


tabuna:
You can inherit any field and add your own list of allowed attributes.
But you're right, this can be a tedious task. 
I think we could automatically allow all attributes starting with `data-`

What do you think about this? This will also solve your problem.

vladitot:
> I think we could automatically allow all attributes starting with `data-`

I can make another fix, and allow to use data- attributes. And I think it is better, really, because you have opportunity to use them, but there no data- attributes in default fields.

Why I dont want to make own fields - because default orchid fields are very good to use, and make own overrides in projects - it is not good way I think.

So, I can just create pull request and change
`protected function getAllowDataAttributes()
    {
        return $this->getAllowAttributes()->filter(function (/* @noinspection PhpUnusedParameterInspection */ $value, $key) {
            return Str::startsWith($key, 'data-');
        });
    }`

to

`protected function getAllowDataAttributes()
    {
        return collect($this->getAttributes())->filter(function (/* @noinspection PhpUnusedParameterInspection */ $value, $key) {
            return Str::startsWith($key, 'data-');
        });
    }`

PR #1240, opened 2020-09-17
Allow data attributes

vladitot:
Request as Decision for https://github.com/orchidsoftware/platform/pull/1232

### DIFF
--- a/src/Screen/Field.php
+++ b/src/Screen/Field.php
@@ -252,7 +252,7 @@ class Field implements Fieldable
      */
     protected function getAllowDataAttributes()
     {
-        return $this->getAllowAttributes()->filter(function (/* @noinspection PhpUnusedParameterInspection */ $value, $key) {
+        return collect($this->getAttributes())->filter(function (/* @noinspection PhpUnusedParameterInspection */ $value, $key) {
             return Str::startsWith($key, 'data-');
         });
     }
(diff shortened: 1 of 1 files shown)


tabuna:
Hey. I think we can write this much better by changing the `getAllowAttributes` method itself.

```php
    /**
     * @return Collection
     */
    protected function getAllowAttributes(): Collection
    {
        $allow = array_merge($this->universalAttributes, $this->inlineAttributes);

        return collect($this->getAttributes())->filter(function ($attribute) use ($allow) {
            return Str::is($allow, $attribute);
        });
    }
```

Something like that. Then we give in `universalAttributes` value `data-*`.
This will fit the pattern and all date attributes can be recorded, but at the same time excluded if necessary.
This also requires confirmation with a test. Tell me if you need help

vladitot:
@tabuna look please, is it right way?

tabuna:
Yes, the meaning is. But this code doesn't work on some `php/laravel` versions, which is signaled by tests.
https://github.com/orchidsoftware/platform/pull/1240/checks?check_run_id=1128384557

tabuna:
Hey @vladitot. Have you looked at the previous post? Do you need help?

tabuna:
https://github.com/orchidsoftware/platform/releases/tag/8.3.0

vladitot:
@tabuna Thank you for your help. There was weekends, so I couldn't finish request.

PR #1242, opened 2020-09-17
Update phpDoc in Orchid\Screen\Concerns\Makeable

marad3r:
phpStan (PHP Static Analysis Tool) throws an error #Call to an undefined method

Fixes #

## Proposed Changes

  -
  -
  -


tabuna:
Hi, I think there should be `static`

tabuna:
Changed in `8.2.1`

marad3r:
I made it like in Screen/Concerns/Multipliable. But you're right that static would be better

PR #1250, opened 2020-09-24
Update de.json

blankster:
* Improved German translation (better German)
* Translated new term
* Removed special char in front of "ß" (special-char German "ss") - see remark below

Remark: In the front of each "ß" (2 times) there was an invisible special car I think could be available by mistake/encoding challenges. I'm not sure if it really was necessary - if it was, please re-add them.


tabuna:
> Remark: In the front of each "ß" (2 times) there was an invisible special car I think could be available by mistake/encoding challenges. I'm not sure if it really was necessary - if it was, please re-add them.

It is very sad to hear that the project contains references toss units. 
I don't think there is any reason to leave this in the project.

PR #1254, opened 2020-09-29
Add custom controls to Quill toolbar

astersnake:
## The problem

is not possible to change the Quill toolbar

## Proposed Changes

  - Give the option to add custom controls to Quill toolbar
  - It allows the user to select 6 groups of controls

## Example

```javascript
Quill::make('quill')
    ->title('Quill')
    ->toolbar(["text", "color", "header", "list", "format", "media"])
```

### DIFF
--- a/resources/js/controllers/fields/quill_controller.js
+++ b/resources/js/controllers/fields/quill_controller.js
@@ -77,20 +77,20 @@ export default class extends Controller {
     }
 
     containerCustomToolbar() {
-        const groups = JSON.parse(this.data.get("toolbar"));
+        const controlsGroup = {
+            "text":   ['bold', 'italic', 'underline', 'strike', 'link', 'clean'],
+            "color":  [{color: this.colors()}, {background: this.colors()}],
+            "header": [{header: '1'}, {header: '2'}],
+            "list":   [{list: 'ordered'}, {list: 'bullet'}],
+            "format": [{indent: '-1'}, {indent: '+1'}, {align: []}],
+            "media":  ['image', 'video'],
+        }
         let container = [];
-        if (groups.includes('text'))
-            container.push(['bold', 'italic', 'underline', 'strike', 'link', 'clean'])
-        if (groups.includes('color'))
-            container.push([{color: this.colors()}, {background: this.colors()}])
-        if (groups.includes('header'))
-            container.push([{header: '1'}, {header: '2'}])
-        if (groups.includes('list'))
-            container.push([{list: 'ordered'}, {list: 'bullet'}])
-        if (groups.includes('format'))
-            container.push([{indent: '-1'}, {indent: '+1'}, {align: []}])
-        if (groups.includes('media'))
-            container.push(['image', 'video'])
+
+        JSON.parse(this.data.get("toolbar")).forEach(function (element) {
+            if (element in controlsGroup)
+                container.push(controlsGroup[element])
+        });
 
         return container
     }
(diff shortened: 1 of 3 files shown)


resources/js/controllers/fields/quill_controller.js
```diff
@@ -76,6 +76,25 @@ export default class extends Controller {
...
+            "media":  ['image', 'video'],
+        }
+        let container = [];
+
+        JSON.parse(this.data.get("toolbar")).forEach(function (element) {
+            if (element in controlsGroup)
+                container.push(controlsGroup[element])
+        });
+
+        return container
```
tabuna, line 95: Hi, I think we can write it down easier:

```javascript
JSON.parse(this.data.get("toolbar")).filter(element => element in controlsGroup);
```
We will not define a new variable and create a second array.
What do you think about this?

resources/js/controllers/fields/quill_controller.js
```diff
@@ -76,6 +76,25 @@ export default class extends Controller {
         ];
     }
 
+    containerCustomToolbar() {
```
tabuna, line 79: What do you think about not duplicating definitions but using a single method named `containerToolbar` ?

src/Screen/Fields/Quill.php
```diff
@@ -41,8 +41,9 @@ class Quill extends Field
      * @var array
      */
     protected $attributes = [
-        'value'  => null,
-        'height' => '300px',
+        'value'   => null,
+        'toolbar' => null,
```
tabuna, line 45: What do you think to list by default all available groups here?

```php
'toolbar' => ["text", "color", "header", "list", "format", "media"]
```

src/Screen/Fields/Quill.php
```diff
@@ -71,4 +72,16 @@ class Quill extends Field
...
     ];
+
+    /**
+     * The given options will be add to the Toolbar.
+     *
+     * @return self
+     */
+    public function toolbar(array $options): self
+    {
+        $this->set('toolbar', json_encode($options));
```
tabuna, line 83: We can define the value using `docblock`. 
And call the method in the constructor so that the value is converted to JSON.

```php
* @method Quill toolbar(array $options)
// ...

    public function __construct()
    {
        $this->addBeforeRender(function () {
                $toolbar = $this->get('toolbar');
                $this->set('toolbar', json_encode($value));
            }
        });
    }

```
This will allow defaults to be specified in the array, and the JSON will only be converted on HTML output.
What do you think about this?

resources/js/controllers/fields/quill_controller.js
astersnake: Yes, this is better

src/Screen/Fields/Quill.php
astersnake: Yes, this is a good idea, I did not think about it.

resources/js/controllers/fields/quill_controller.js
astersnake: hmmm, and use [this](https://github.com/orchidsoftware/platform/pull/1254#discussion_r496589195) as default?

src/Screen/Fields/Quill.php
astersnake: ok, this is beyond my knowledge. But if you say is better, must be.

resources/js/controllers/fields/quill_controller.js
tabuna: Yes, when we define the default values, we won't need to duplicate them.

resources/js/controllers/fields/quill_controller.js
astersnake: Ok, I ended up with this:
```js
return JSON.parse(this.data.get("toolbar"))
            .map(tool => controlsGroup[tool]);
```

PR #1263, opened 2020-10-01
ckeditor

timeshift92:
Fixes #

## Proposed Changes

  - add new field
  -
  -


tabuna:
Hey @timeshift92.

Thanks for your work :muscle: . I have no plans to add new wysing editors right now. Since they significantly increase the maintenance costs and the amount of JS/CSS received.

But please publish this as a separate package. If you need help with this, you can count on my support. For people to know about it, we can also add a link to it in the [awesome repository](https://github.com/orchidsoftware/awesome)

Issue #1273, opened 2020-10-05
User List: Column "id" doesn't contain the user id (wrong label or content)


blankster:
**Describe the bug**
In the listing of all available users (/users) the last column is labelled with "id" but is containing function for edit and delete. A user/admin will expect to see the id of the user and/or another label (e.g. "operation" or "options").

**To Reproduce**

  1. Setup a basic installation of orchid
  2. Execute a seeder that creates users
  3. Login as admin
  4. Access the user list (path: /users)

**Expected behavior**
Display the user id and/or the column for operations/options. The label of the column is misleading as the column doesn't contain the id.

**Screenshots**
![image](https://user-images.githubusercontent.com/1443109/95133455-a23c1200-0761-11eb-8489-e38896167dc4.png)

Let me know if you need environment-specific information but I guess this little bug is occurring environment independent in the currently latest release.

PR #1289, opened 2020-10-09
Corrected Users listing heading label. Changed heading label from ID …

luckys383:
Corrected Users listing heading label. Changed heading label from ID to Actions

Fixes #
  - User List: Column "id" doesn't contain the user id (wrong label or content)

Issue #1273

### DIFF
--- a/stubs/app/Orchid/Layouts/User/UserListLayout.php
+++ b/stubs/app/Orchid/Layouts/User/UserListLayout.php
@@ -54,7 +54,7 @@ class UserListLayout extends Table
                     return $user->updated_at->toDateTimeString();
                 }),
 
-            TD::set('id', 'ID')
+            TD::set('actions', 'Actions')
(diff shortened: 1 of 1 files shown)


stubs/app/Orchid/Layouts/User/UserListLayout.php
```diff
@@ -54,7 +54,7 @@ public function columns(): array
                     return $user->updated_at->toDateTimeString();
                 }),
 
-            TD::set('id', 'ID')
+            TD::set('actions', 'Actions')
```
tabuna, line 57: Hey. There is no value binding, which means we can specify a record without a second argument.
```php
TD::set(__('Actions'))
```
It would also be great if we wrapped the string as translation

luckys383: @tabuna added string to the translation.

Issue #1276, opened 2020-10-06
btn class not working on Link


RaiserWeb:
Using Link::make()->set('class', 'btn btn-default') does add the class to the `<a>` tag in the html, but the style is being overridden by other css, so it's not displaying. It would be great for this style to work on Links

RaiserWeb:
For more info, this happens when the link is inside a table layout

PR #1291, opened 2020-10-10
#1276: btn class not working on Link

luckys383:
Fixes #
btn class style not working inside table.

issue #1276


luckys383:
@tabuna could you review and merge this PR, if everything is correct?

tabuna:
Hi, I have no implementation questions.
But I'm not sure if this is what we need.

I intentionally disable button styles in `commandbar` and `tables`.
So that there are no large non-natural indents when text goes to these buttons:

```php
TD::set('name', __('Name'))
    ->sort()
    ->cantHide()
    ->filter(TD::FILTER_TEXT)
    ->render(function (Role $role) {
        return Link::make($role->name)
            ->route('platform.systems.roles.edit', $role->id);
    }),
````



The only thing that doesn't suit me is that the color of the links and buttons is the same as the normal color:

image

For example, here, it is not very clear where you can click and where not.


But let's try to accept this.

PR #1297, opened 2020-10-12
added en.js

Yarynka:
Fixes #

## Proposed Changes

  -
  -
  -


tabuna:
Hey @Yarynka 

Could you clarify what this change request?
The presence of such a file might be useful for creating new translations. But it has different meanings from the current ones. If you think that the current selection of phrases is bad, then let's make changes first. And then, we will create a template file for future translations.

PR #1302, opened 2020-10-14
Update Screen.php

vladitot:
Incorrect parameters substitution in method like "remove" at UserScreen

### DIFF
--- a/src/Screen/Screen.php
+++ b/src/Screen/Screen.php
@@ -168,7 +168,7 @@ abstract class Screen extends Controller
         $query = ! is_array($query) ? [] : $query;
 
         $parameters = array_filter($parameters);
-        $parameters = array_merge($query, $parameters);
+        $parameters = array_merge($parameters, $query);
 
         $response = $this->callMethod($method, $parameters);
 


vladitot:
@tabuna hi! I don't really know why no-one more noticed this bug.
But there is a problem, when you have Button::method('remove') and public function remove(User $user) 
User is with empty attributes. That is because first argument on callMethod is "route url". But it should be last argument. I think it was in platform 7.*
Or maybe not. So, please, review this request and emit patch please.
Have a good day!

tabuna:
Hey @vladitot.
 I am not experiencing any problem with deleting users. 
How can I reproduce your problem?

vladitot:
I will try show receipt in few hours.

vladitot:
I'm glad sorry. I found what I lost. My nginx was set up the way, an additional parameter added on each query. Laravel thought that there is one more query parameter, and route parameters were mapping incorrectly. I think my fix in this pull request can save people in same situation, but if there is no same situations - just close my pull request.

tabuna:
I don't think it needs to be changed now. 
But I think it would be great to describe the problem and its solution for those who will face it in the future.

PR #1313, opened 2020-10-19
Fixed display of values with double quotes in text inputs

Altair4713:
Values with double quotes were displayed correctly in text inputs.

Example:
image

### DIFF
--- a/src/Screen/Fields/Input.php
+++ b/src/Screen/Fields/Input.php
@@ -103,6 +103,10 @@ class Input extends Field
             if (is_array($mask)) {
                 $this->set('mask', json_encode($mask));
             }
+
+		        if(!empty($this->get('value'))) {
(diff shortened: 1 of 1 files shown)


tabuna:
Made the fixes myself. The proposed changes do not affect all attribute behaviors. https://github.com/orchidsoftware/platform/pull/1314

In previous versions, we used our own attribute handling. A newer version of Laravel has a class `ComponentAttributeBag`.
Which is created for the same purposes. It turned out that the data passed to the constructor is not processed:

```php
$bag = new ComponentAttributeBag(['value' => 'valueQuote"']);

(string) $bag // value="valueQuote""
```

Used instead:

```php
$bag = new ComponentAttributeBag();
$bag->merge(['value' => 'valueQuote"']);

(string) $bag // value="valueQuote&quot;"
```

PR #1346, opened 2020-11-09
Added indication of the label name

Vladqqw:
Added instructions for the name label, not all use translation by key, `resources/lang/{lang}.json`, violates the integrity of the project if there are a lot of languages in the project
`resources/lang/{lang}/platform.php`

```
 Map::make('place')
                    ->latitude(_('admin.map.latitude'))
                    ->longitude(_('admin.map.longitude'))
                    ->title('Object on the map')
                    ->help('Enter the coordinates, or use the search')
```

### DIFF
--- a/resources/views/fields/map.blade.php
+++ b/resources/views/fields/map.blade.php
@@ -8,7 +8,7 @@
         </div>
         <div class="row mt-3">
             <div class="col-md">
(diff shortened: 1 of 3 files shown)


tabuna:
Hey. You can specify translations of these keys directly in this package by making a PR.
This will help everyone because there will be no need to create files with translations for each developer independently.

PR #1350, opened 2020-11-10
ref #1349 [Upload] Fix upload timeout

### LINKED ISSUE #1349: Dropzone.js timeout option not set. Using 30 seconds default.
**Is your feature request related to a problem? Please describe.**
When dealing with large files with the Upload field, the default Dropzone.js option of timeout is set to drop upload after 30 seconds (which matches the default of PHP).

**Describe the solution you'd like**
It would be great if it would take into account the php max_execution_time for this option. Or to customize it via a "timeout" method on the field.

**Describe alternatives you've considered**
- Add "max_execution_time" to "maxFileUpload" [method](https://github.com/orchidsoftware/platform/blob/03f9a113b1a70bc5075ce86a918707f0e7d82169/src/Support/Init.php#L60) and add the option on [initDropZone](https://github.com/orchidsoftware/platform/blob/6d1520d0bdf879fefc45c300084a9c6b0620f7bf/resources/js/controllers/fields/upload_controller.js#L190) calling the "max_execution_time" option from Init.
- Just set a default (0) and giving one the ability to change the option from the Field.

**Additional context**
Wanted to create a PR but decided to create an issue to expose the problem, since not sure if mine is the best solution (by architecture). But happy to submit a PR.

josespinal:
Fixes #1349

## Proposed Changes

  - Add a timeout option to the Upload field with default of 0.
  - Add the Dropzone option of "timeout" and get it from the Field definition.


src/Screen/Fields/Upload.php
```diff
@@ -56,6 +56,7 @@ class Upload extends Field
         'parallelUploads' => 10,
         'maxFileSize'     => null,
         'maxFiles'        => 9999,
+        'timeOut'         => 0,
```
tabuna, line 59: Let's add a dock block with an explanation that this value can be overridden:

```php
* @method Upload timeOut(int $second = null)
```

PR #1377, opened 2020-11-27
Search columns for Relation field

grigoriy-ivanov:
Multiple columns for search in Relation field

```php
Relation::make('customer_id')
    ->fromModel(Customer::class, 'name', 'id')
    ->searchColumns('company','tin')
```

### DIFF
--- a/resources/js/controllers/fields/relation_controller.js
+++ b/resources/js/controllers/fields/relation_controller.js
@@ -16,6 +16,7 @@ export default class extends Controller {
         const key = this.data.get('key');
         const scope = this.data.get('scope');
         const append = this.data.get('append');
+        const columns = this.data.get('columns');
 
 
         $.ajaxSetup({
@@ -57,6 +58,7 @@ export default class extends Controller {
                     key,
                     scope,
                     append,
+                    columns,
                 }),
             },
             placeholder: select.getAttribute('placeholder') || '',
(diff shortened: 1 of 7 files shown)


resources/js/controllers/fields/relation_controller.js
```diff
@@ -16,6 +16,7 @@ export default class extends Controller {
         const key = this.data.get('key');
         const scope = this.data.get('scope');
         const append = this.data.get('append');
+        const columns = this.data.get('columns');
```
tabuna, line 19: I think the name `columns` does not convey the whole meaning.
Can we reimagine them to `searchColumns`?

src/Screen/Fields/Relation.php
```diff
@@ -192,6 +194,18 @@ public function applyScope(string $scope, ...$parameters): self
         return $this;
     }
 
+    /**
+     * @param array $columns
+     *
+     * @return $this
+     */
+    public function searchIn(array $columns): self
```
tabuna, line 202: It may not always be required to specify a lot of fields,
 let's change it to `string|array` and wrap the parameter in `Arr::wrap`?

tabuna:
Can you add tests?

tabuna (commented):
Can you add tests?

src/Screen/Fields/Relation.php
grigoriy-ivanov: Maybe?
```php
public function searchIn(...$columns): self

tabuna: Yes, that's okay too

resources/js/controllers/fields/relation_controller.js
grigoriy-ivanov: I rename method ```searchIn``` to ```searchColumns``` too.

Issue #1393, opened 2020-12-10
Translation for persian language farsi


wikiccu:
Hello
I would ask you is there any plan for translation to the Persian/farsi language ?

tabuna:
Hey @wikiccu, I have no plans to add languages that I do not speak.
 But I will gladly accept PR for adding them to the repository.

wikiccu:
Sure I asked this because I wanted to send PR for translation 
thank you ill work on it.

PR #1394, opened 2020-12-11
Create fa.json

wikiccu:
#1393 
## Proposed Changes
create Persian Farsi translation.


tabuna:
Maybe I'm wrong. But isn't Persian written from right to left? Don't you have a problem with that? We could add HTML to tell browsers to change direction.

```html
<html dir="rtl" lang="fa">
```

wikiccu:
> Maybe I'm wrong. But isn't Persian written from right to left? Don't you have a problem with that? We could add HTML to tell browsers to change direction.
> 
> ```
> <html dir="rtl" lang="fa">
> ```

Yes, Persian like Arabic is Right to Left some time it needs CSS changes but for reading in a good way that HTML code can help.
thank you

PR #1470, opened 2020-12-22
Add model key to config/platform.php for overriding the authentication models

level39:
This allows us to easily reference the custom User and Role models if necessary.


tabuna:
Hey, have you seen this https://orchid.software/en/docs/configuration/#model-classes?

level39:
Great, guess there's no need to merge. Thanks!

PR #1486, opened 2020-12-31
Fix datepicker issue on filter dropdown

### LINKED ISSUE #1386: Datepicker month click inside Filter hides filter drop down
**Describe the bug**
When using a DateTimer picker, inside a Orchid\Filters\Filter class, if a user clicks 'back' or 'forward' month arrows, the filter drop down hides, along with the 'Apply' button, so a user needs to re-click 'Filter' drop down to see the apply button.

**Expected behavior**
Filter drop down should remain open.

**Screenshots**
Click on red box:
image

Drop down is hidden:
image

RaiserWeb:
Fixes #1386

## Proposed Changes

- FlatDatePicker config change to set `static: true`
- `.flatpickr-wrapper` class given 100% width
- date filter markup changed to include js stoppropogation to stop the datepicker closing when next/back month buttons are clicked

### DIFF
--- a/resources/views/partials/filters/date.blade.php
+++ b/resources/views/partials/filters/date.blade.php
@@ -1,4 +1,4 @@
-<div class="dropdown d-inline-block">
+<div class="dropdown d-inline-block" data-controller="screen--filter">
     <button class="btn btn-sm btn-link dropdown-toggle"
             type="button"
             data-toggle="dropdown"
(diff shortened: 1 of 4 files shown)


resources/views/partials/filters/date.blade.php
```diff
@@ -1,4 +1,4 @@
-<div class="dropdown d-inline-block">
+<div class="dropdown d-inline-block" data-controller="screen--filter">
```
tabuna, line 1: We need to specify and use this controller in other templates that relate to the filter.

PR #1495, opened 2021-01-10
Add missing Spanish translation

astersnake:
## Proposed Changes

  - Add new Spanish translations


tabuna:
Hi, [some lines are missing](https://github.com/orchidsoftware/platform/pull/1495/checks?check_run_id=1677674830#step:6:20). Accordingly, their translation will not be shown, for example:

> The string 'Basic information' is not used in the project!

astersnake:
Hi, yeah, I Just noticed I was doing the translation using a previous version of Orchid. Will fix it asap.

tabuna:
I'm sorry. But I wouldn't say I liked the past default messages.
 The current ones are a little better, but I think it could be even better;)

astersnake:
No problem is much better now. I already update the file.

PR #1533, opened 2021-01-21
Hide file download link for non-public storage

### LINKED ISSUE #1532: Uploading file to non-public storage using Field Upload gives 404 on Link to File
**Describe the bug**

Switching Field 

```
Upload::make()->storage('local_private')
```

and filesystem.php

```
'local_private' => [
            'driver' => 'local',
            'root'   => storage_path('app/private'),
        ]
```

Gives a Link to File which 404s:

image

image

Thinking:

`/resources/views/fields/upload.blade.php`

```
@if($storage === 'public')
<div class="form-group">
    <a href="#" data-action="click->fields--upload#openLink">
        <small>
            <x-orchid-icon path="link" class="mr-2"/>

            {{ __('Link to file') }}
        </small>
    </a>
</div>
@endif
```

Should work?

Caffe1neAdd1ct:
Fixes #1532 

May be better to check filesystem storage properties for [url|visibility]

```
'url' => env('APP_URL').'/storage',
'visibility' => 'public',
```

Not sure if this is used for S3 so feel free to decline merge if this breaks anything and i'll take another look. Another solution posted in the issue.

### DIFF
--- a/resources/views/fields/upload.blade.php
+++ b/resources/views/fields/upload.blade.php
@@ -85,6 +85,7 @@
                                 </div>
 
 
+                                @if($visibility === 'public')
                                 <div class="form-group">
                                     <a href="#" data-action="click->fields--upload#openLink">
                                         <small>
@@ -94,6 +95,7 @@
                                         </small>
                                     </a>
                                 </div>
+                                @endif
 
 
                             </div>
--- a/src/Screen/Fields/Upload.php
+++ b/src/Screen/Fields/Upload.php
@@ -24,7 +24,6 @@ use Orchid\Support\Init;
  * @method Upload placeholder(string $value = null)
  * @method Upload value($value = true)
  * @method Upload help(string $value = null)
- * @method Upload storage($value = true)
  * @method Upload parallelUploads($value = true)
  * @method Upload maxFileSize($value = true)
  * @method Upload maxFiles($value = true)
@@ -64,6 +63,7 @@ class Upload extends Field
         'resizeHeight'    => null,
         'media'           => false,
         'closeOnAdd'      => false,
+        'visibility'      => 'public',
     ];
 
     /**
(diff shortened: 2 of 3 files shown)


Caffe1neAdd1ct:
Would need to check $storage has actually been set on the upload, default is public. 

See https://github.com/orchidsoftware/platform/issues/1532#issuecomment-765015369

Caffe1neAdd1ct:
Passing tests now, however i'm thinking the "visibility" for the Upload Field should be passed in not calculated in the view.

resources/views/fields/upload.blade.php
```diff
@@ -85,7 +85,7 @@
                                 </div>
 
 
-                                @if($storage === 'public')
+                                @if(config('filesystems')['disks'][$storage ?? 'public']['visibility'] ?? 'private' === 'public')
```
tabuna, line 88: I think we should move this to the field class. And then only use the result.

```php
    /**
     * All attributes that are available to the field.
     *
     * @var array
     */
    protected $attributes = [
        // ...
        'visibility'  => 'public',
    ];
```

Then we would only check small nesting in view files:

```blade
@if($visibility === 'public')
   {{--  code --}}
@endif
```

And they would check already in the method:

```php
/**
 * @param string $storage
 *
 * @return $this
 * @throws \Throwable
 */
public function storage(string $storage) : self
{
    $disks = config('filesystems.disks');

    throw_unless(
        array_key_exists($storage, $disks),
        \Exception::class,
        'The selected storage was not found'
    );

    $this->set('storage', $storage);
    $this->set('visibility', Arr::get($disks, "$storage.visibility"));
}
```

What do you think about it?

tabuna: Or you can write it even shorter.

```php
public function storage(string $storage) : self
{
    $disk = config("filesystems.disks.$storage");

    throw_if(
        $storage === null,
        \Exception::class,
        'The selected storage was not found'
    );

    return $this
        ->set('storage', $storage)
        ->set('visibility', $disk['visibility'] ?? null);
}
```

Caffe1neAdd1ct: That's a much better solution, love it :+1: 

Would you like me to implement or are you happy to change the pull request?

tabuna: Please prepare a PR, with such a solution, as well as tests. If you have any difficulties, I will help you.

Caffe1neAdd1ct: @tabuna Will put on tonight todo list :+1: 

Thanks for the pointers, very helpful and good to start understanding how the platform works.

Caffe1neAdd1ct:
7.3 issue fixed in phpunit test, changed `(string) $upload` to `$upload->render()`

Caffe1neAdd1ct:
image

Code coverage improved to 96% 
Unit tests implemented to cover storage function for visibility
PHP 7.3 __toString exception caused in new unit test fixed
The two issues identified by phan look like no issue

Issue #1527, opened 2021-01-18
Create new user via Systems -> Users screen


pqr:
**Is your feature request related to a problem? Please describe.**
With default Orchid installation there is no possibility to create a new user via UI (on Systems -> Users screen), there is no "Add user" or "Create user" button.

I think it would be nice to have an option to create users from UI with default Orchid installation.

**Describe the solution you'd like**
I want to add "+ Add" button on Users screen.

One important addition: since password field in database have no default value I propose to add a "Password" field on user create/edit screen.
- When Creating a new user the Password field will be mandatory to fill in. This to prevent user creation with empty password by accident.
- When editing existing user the Password field might be left blank - means do not change current password.

Let us discuss this proposal. If agreed, I will implement and make a Pull Request.

tabuna:
Hi, do you want a separate page with roles and permissions or a simple modal with email and password?

pqr:
<img width="1190" alt="image" src="https://user-images.githubusercontent.com/451007/104905667-bb28a600-5993-11eb-939f-7e33f11aac32.png">


<img width="1174" alt="image" src="https://user-images.githubusercontent.com/451007/104905597-a5b37c00-5993-11eb-82fc-bc77c394893e.png">

dhb52:
`Role/User` data management together into a `Role and User` Menu group.

PR #1534, opened 2021-01-22
Create new user via Systems -> Users screen

pqr:
This is implementation for issue #1527

Add ability to create new user via Users admin screen.

### DIFF
--- a/stubs/app/Orchid/Screens/User/UserEditScreen.php
+++ b/stubs/app/Orchid/Screens/User/UserEditScreen.php
@@ -60,7 +60,7 @@ class UserEditScreen extends Screen
         $user->load(['roles']);
 
         return [
-            'user' => $user,
+            'user'       => $user,
             'permission' => $user->getStatusPermission(),
         ];
     }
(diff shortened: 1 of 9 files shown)


stubs/app/Orchid/Screens/User/UserListScreen.php
```diff
@@ -57,7 +58,11 @@ public function query(): array
      */
     public function commandBar(): array
     {
-        return [];
+        return [
+            Link::make(__('Add'))
+                ->icon('plus')
+                ->href(route('platform.systems.users.create')),
```
tabuna, line 64: We can use a shorter notation:


```php
Link::make(__('Add'))
    ->icon('plus')
    ->route('platform.systems.users.create'),
```

stubs/app/Orchid/Screens/User/UserEditScreen.php
```diff
@@ -150,14 +162,23 @@ public function save(User $user, Request $request)
             ->collapse()
             ->toArray();
 
+        $userData = $request->get('user');
```
tabuna, line 165: It looks like the $ userData variable is not being used. I think you meant it?

```php
$password = $request->input('user.password');

if (!$user->exists && !empty($password)) {
    $user->password = Hash::make($password);
}
```

pqr: My fail with last minute changes.
Fixed

stubs/app/Orchid/Screens/User/UserListScreen.php
pqr: Nice addition, thanks!

PR #1536, opened 2021-01-23
Make UserEditScreen stub more reasonable for CREATE

dhb52:
Fixes #

UserEditScreen stub more reasonable for CREATE

## Proposed Changes

  - When creating a new user, `Remove` and `LoginAs` button is not available.
  - When creating a new user, `Save` button is not available for blocks.
  - Screen name `Edit User` for Editing, `Create User` for Creating.


stubs/app/Orchid/Screens/User/UserEditScreen.php
```diff
@@ -56,6 +56,11 @@ class UserEditScreen extends Screen
     public function query(User $user): array
     {
         $this->user = $user;
+        $this->exists = $user->exists;
+
+        if (! $this->exists) {
+            $this->name = 'Create User';
+        }
```
dhb52, line 63: Is it better?
```php
$this->editing = $user->exists;
```

PR #1547, opened 2021-02-01
Add a languages scan and sync tool

dhb52:
# Add a languages scan and sync tool

## Proposed Changes

  - scan `src, stubs, resources/views` for `__()`
  - scan translatable field from XxxScreen,  `$name`, `$description`
  - loads resources/lang/*.json
  - order translations by key
  - when existing key not included in some json translation, `[*].$key` is for translation, which indicates further translation needed.

### DIFF
--- a/lang.php
+++ b/lang.php
@@ -3,9 +3,9 @@
 # adapted from https://github.com/barryvdh/laravel-translation-manager
 function find_translation_keys(): array
 {
-    $files = glob_recursive("./src", "*");
-    $files = array_merge(glob_recursive("./stubs", "*"), $files);
-    $files = array_merge(glob_recursive("./resources/views", "*"), $files);
+    $files = glob_recursive("./src", "*.php");
+    $files = array_merge(glob_recursive("./stubs", "*.php"), $files);
+    $files = array_merge(glob_recursive("./resources/views", "*.php"), $files);
     $files = array_filter($files, function ($f) {
         return is_file($f);
     });
@@ -24,6 +24,19 @@ function find_translation_keys(): array
         "\s*[\),]";                                     // Close parentheses or new parameter
 
 
+    $translatables = [
+        "name",
+        "description",
+    ];
+    $translatablePattern =                              // public $name = "Create User"; $this->name = "Edit User";
+        '(public[^\w]*\$|\$this->)'.                    // public | $this->
+        '('.implode('|', $translatables).')'.           // (name|description)
+        "[^\w]*=[^\w]*".                                // Blank
+        "(?P<quote>['\"])".                             // Match " or ' and store in {quote}
+        "(?P<string>(?:\\\k{quote}|(?!\k{quote}).)*)".  // Match any string that can be {quote} escaped
+        "\k{quote}".                                    // Match " or ' previously matched
+        "[^\w]*;";                                            // End with ";"
+
     $keys = [];
 
     foreach ($files as $file) {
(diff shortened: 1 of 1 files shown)


dhb52:
`php lang.php > summary.md`

tabuna:
Hi @dhb52, very cool idea, but what do you think would be a separate package that the class will use?
We would connect it to GitHub Actions, and it would do it automatically.

For example:

```bash
php vendor/bin/lang.php --test
```
Check and issue warnings.


And for development, we would use the bin dependency, for example
```bash
php vendor/bin/lang.php --fix
```

Which would have already added the missing translation.

I can help you with this if you want.

dhb52:
> Hi @dhb52, very cool idea, but what do you think would be a separate package that the class will use?
> We would connect it to GitHub Actions, and it would do it automatically.
> 
> For example:
> 
> ```shell
> php vendor/bin/lang.php --test
> ```
> 
> Check and issue warnings.
> 
> And for development, we would use the bin dependency, for example
> 
> ```shell
> php vendor/bin/lang.php --fix
> ```
> 
> Which would have already added the missing translation.
> 
> I can help you with this if you want.

Thanks for offering help, my php skill is quite poor.
Originally, I tried writing this tool without bringing in `composer`. 
And what I've done is for `--fix`, 
and I need to do dry-run as `--test`, which is quite helpful to notify developer when adding new `translation` into the commited code.

As I put `[*].$key` for missing translation, today, I also wrote a python tool for cloud translation provided by `Baidu`, and the code goes like this:

```python
# coding=utf-8

import glob
import hashlib
import http.client
import json
import os.path
import random
import urllib
import urllib.parse
import time


class BaiduTranslator:
    host = "api.fanyi.baidu.com"
    uri = "/api/trans/vip/translate"

    def __init__(self, appid, secret, to_lang, from_lang="en") -> None:
        self.appid = appid
        self.secret = secret
        self.to_lang = to_lang
        self.from_lang = from_lang

    def translate(self, q):
        http_client = None

        salt = random.randint(32768, 65536)
        sign = self.appid + q + str(salt) + self.secret
        sign = hashlib.md5(sign.encode()).hexdigest()

        params = {
            "appid": self.appid,
            "sign": sign,
            "from": self.from_lang,
            "q": q,
            "to": self.to_lang,
            "salt": salt,
            "sign": sign,
        }

        data = urllib.parse.urlencode(params)
        url = f"{self.uri}?{data}"

        result = None
        try:
            http_client = http.client.HTTPConnection(self.host)
            http_client.request("GET", url)
            response = http_client.getresponse()
            result_all = response.read().decode("utf-8")
            result = json.loads(result_all)
        except Exception as e:
            print(e)
        finally:
            if http_client:
                http_client.close()

        return result["trans_result"][0]["dst"] if result != None else None


if __name__ == "__main__":
    for path in glob.glob("resources/lang/*.json"):
        file_name = os.path.basename(path)
        locale = file_name[: -len(".json")]
        baidu = BaiduTranslator(
            appid="YOUR APPID", secret="YOUR SECRET", to_lang=locale
        )
        with open(path) as json_file:
            translation = json.load(json_file)
            for (key, val) in translation.items():
                if val.startswith("[*]"):
                    new_val = baidu.translate(key)
                    time.sleep(1)
                    translation[key] = new_val
            print(translation) #TODO write back to file.
```
With these two tools, we can't quickly creating a new `machine translated` language file quickly.
And I think these auto tool should be run manually.

BTW: I would like to use Google Cloud Translator at first, but it's not available for China. ;-(

dhb52:
a composer package is here: https://github.com/dhb52/orchid-lang

PR #1551, opened 2021-02-06
language file normalized by dhb52/orchid-lang tools

dhb52:
fa.json, uk.json NOT translated by BAIDU.


resources/lang/fa.json
```diff
@@ -1,103 +1,154 @@
...
     "Sign out": "خروج",
-    "Size": "اندازه",
     "Slug": "اسلاگ (Slug)",
-    "Status": "وضعیت",
-    "System name": "نام سیستم",
     "System": "سیستم",
+    "System name": "نام سیستم",
     "System settings": "تنظیمات سیستم",
-    "URL": "آدرس اینترنتی",
+    "Systems": "[*]Systems",
```
tabuna, line 131: Hi. Are there any reasons why we should show a `[*]` prefixed string?

blankster: Next to this the plural "systems" is used, but it should be the singular "system". This was recently changed in any languages.

blankster: @tabuna this is also still wrong in Russian.

dhb52: There are two steps to produce this PR:
Step 1: scan the source code (src/stubs/views) for `__()` translation function call, which generate the set of translation key string. then look up the original json files for existing translations, if not found, `[*]$key` will be the temporary translation.

Step 2: A machine translation BAIDU is called in to help translate the `temporary translation` from step 1, if the translation is not available from BAIDU, then `temporary translation` will remain, `[*]` indicates  human translation is needed.

As I mentioned, `fa.json` and `uk.json` is not available from BAIDU.

Issue #1609, opened 2021-03-13
[10.0] Lack of  (or incomplete) implicit model binding on screen methods


lemyskaman:
**Is your feature request related to a problem? Please describe.**

On laravel controllers methods, as long as I pass a formrequest and a model as type-hint params (with the same var name as the route defined param) at the method heading, laravel automatically binds the model and made it becomes available from the formrequest route method (using the same route attribute name). If I'm not wrong that's what is called Implicit Binding on laravel.

When using Orchid's Screens and I try to replicate that behavior on any screen method I got from $request->route('param') just the value of the param (model key) instead of the hydrated model.

To illustrate the thing better let's say I have:

- a model (SomeModel)
```php
//app/Models/SomeModel.php
use Illuminate\Database\Eloquent\Model;

class SomeModel extends Model
{
    use HasFactory;
}

```

- a form request (SomeModelFormRequest)
```php
//app/Http/Request/SomModelRequest.php

use Illuminate\Foundation\Http\FormRequest;

class SomeModelFormRequest extends FormRequest
{
    public function authorize()
    {
        return true;
    }

    public function rules()
    {
    
        return [
            //
        ];
    }
}
```

- a screen (SomeScreen) 
```php
//App/Screens/SomeScreen.php

use Orchid\Screen\Screen;

class SomeScreen extends Screen
{

    public $name = 'OrchidTestScreen';

    public $description = 'OrchidTestScreen';

    public function query(SomeModel $somemodel,SomeModelFormRequest $request): array
    {

        dd($request->route('somemodel'));
        return [];
    }

```
- and a controller (SomeController) to compare with the SomeScreen binding behavior.
```php
class SomeControler extends Controller
{
    //
    //
    public function index(SomeModel $somemodel,SomeModelFormRequest $request){

    	dd($request->route('somemodel'));

    }
}

```

If then we drive request with this routes at a route file:

```php

Route::get('plaincontroller/{somemodel?}/edit',[App\Http\Controllers\SomeControler::class,'index'])
    ->name('platform.controller');

Route::screen('/plainscreen/{somemodel?}/edit',App\Screens\SomeScreen::class)
    ->name('platform.screen');

```

Then pointing the browser at "platform.screen" route will display the value you passed on the get request from the browser as "somemodel" param, when instead oone would expect to get the model with that primary key attribute value as would happen if we point the browser to the "platform.controller" route.

**Describe alternatives you've considered**
[...]

PR #1610, opened 2021-03-14
improved route binding, with dinamic model binding on route according…

lemyskaman:
… issue #1609

Fixes #

## Proposed Changes

  - Set the model on the moment of screen binding method
  -
  -


src/Screen/Screen.php
```diff
@@ -236,7 +236,11 @@ private function bind(int $key, ReflectionParameter $parameter, array $httpQuery
         $object = resolve($class);
 
         if ($original !== null && is_a($object, UrlRoutable::class)) {
-            return $object->resolveRouteBinding($original);
+            $object_to_return = $object->resolveRouteBinding($original);
```
tabuna, line 239: Variables are written in the style `$objectToReturn`.  By the way, I don't think such a name is appropriate.

src/Screen/Screen.php
```diff
@@ -236,7 +236,11 @@ private function bind(int $key, ReflectionParameter $parameter, array $httpQuery
...
+            if (key_exists($parameter->getName(),request()->route()->parameters)){
```
tabuna, line 240: Are you sure we need this check? Don't we do the same a few levels higher.

lemyskaman: Well in order to replicate normal controllers behavior on binding:

   When you pass a typehint FormRequest object on a controller method as param, the request only has the value of a resource/model in the route attribute,
   
   But if you pass the model, as typehint object and also a formrequest in the controller method heading, then the route attribute value is replaced by the model with data from the db that match the old route attribute value as id.
   
   I don't see that in upper levels.

lemyskaman: off course the route  attribute name, and the controller method typedhint object name as parameter must match to do the route attribute replacement

src/Screen/Screen.php
lemyskaman: i agree to change the name of the var,

  dint know var can only be written in camelcase.

src/Screen/Screen.php
lemyskaman: You are right, Check is not needed

src/Screen/Screen.php
lemyskaman: what about, $boundObject ?   as it make more sense

tabuna: What do you think about changing the early exit from the function so that it would be like this:

```php
$object = resolve($class);

if ($original === null || !is_a($object, UrlRoutable::class)) {
    return $object;
}

return tap($object->resolveRouteBinding($original), function ($object) use ($parameter) {
    Route::current()->setParameter($parameter->getName(), $object);
});
```

tabuna:
[Released 10.0.0](https://github.com/orchidsoftware/platform/releases/tag/10.0.0)

PR #1614, opened 2021-03-17
Added multiguard support. Allows Orchid to be use with a custom auth …

ShamarKellman:
Add Multiguard support.
This pull request allows for the use of multiple guards. In a case where there are separate models for admin and frontend. 
If the custom guard is specified the user will be able to authenticate admin uses separately from regular users.

Fixes #

## Proposed Changes

  - Added guard config to config file, uses default guard if no custom guard is specified.
  - Use guard method when retrieving authenticated user

### DIFF
--- a/config/platform.php
+++ b/config/platform.php
@@ -54,6 +54,22 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Guard
+    |--------------------------------------------------------------------------
+    |
+    | This can be used if you are using a multi-auth set up configured.
+    | Such as using a Admin model for Orchid and User Model for frontend access.
+    | If not using default auth guard remember to add 'auth:guard_name' to the middleware
+    | where guard_name is the custom guard name.
+    |
+    | You can learn more here: https://laravel.com/docs/authentication
+    |
+    */
+
+    'guard' => config('auth.defaults.guard'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Auth Page
     |--------------------------------------------------------------------------
     |
--- a/resources/views/app.blade.php
+++ b/resources/views/app.blade.php
@@ -11,7 +11,7 @@
         @endif
     </title>
     <meta name="csrf_token" content="{{  csrf_token() }}" id="csrf_token" data-turbolinks-permanent>
-    <meta name="auth" content="{{  Auth::check() }}" id="auth" data-turbolinks-permanent>
+    <meta name="auth" content="{{  Auth::guard(config('platform.guard'))->check() }}" id="auth" data-turbolinks-permanent>
     @if(file_exists(public_path('/css/orchid/orchid.css')))
         <link rel="stylesheet" type="text/css" href="{{  mix('/css/orchid/orchid.css') }}">
     @else
--- a/resources/views/dashboard.blade.php
+++ b/resources/views/dashboard.blade.php
@@ -23,7 +23,7 @@
 
                 @include('platform::partials.search')
 
-                @includeWhen(Auth::check(), 'platform::partials.profile')
+                @includeWhen(Auth::guard(config('platform.guard'))->check(), 'platform::partials.profile')
 
                 <ul class="nav flex-column mb-1">
                     {!! Dashboard::menu()->render('Main') !!}
--- a/resources/views/partials/profile.blade.php
+++ b/resources/views/partials/profile.blade.php
@@ -1,14 +1,14 @@
 <div class="row m-0 align-items-center p-3">
     <div class="dropdown col p-0">
         <a href="#" class="nav-link p-0 v-center" data-toggle="dropdown">
-            @if($image = Auth::user()->presenter()->image())
+            @if($image = Auth::guard(config('platform.guard'))->user()->presenter()->image())
                 <span class="thumb-sm avatar mr-3">
                         <img src="{{$image}}" class="b">
                 </span>
             @endif
             <span style="width:12em;font-size: 0.85em;">
-                <span class="text-ellipsis">{{Auth::user()->presenter()->title()}}</span>
-                <span class="text-muted d-block text-ellipsis">{{Auth::user()->presenter()->subTitle()}}</span>
+                <span class="text-ellipsis">{{Auth::guard(config('platform.guard'))->user()->presenter()->title()}}</span>
+                <span class="text-muted d-block text-ellipsis">{{Auth::guard(config('platform.guard'))->user()->presenter()->subTitle()}}</span>
             </span>
         </a>
         <div class="dropdown-menu dropdown-menu-left dropdown-menu-arrow bg-white">
@@ -19,7 +19,7 @@
                 <div class="dropdown-divider"></div>
             @endif
 
-            @if(Auth::user()->hasAccess('platform.systems.index'))
+            @if(Auth::guard(config('platform.guard'))->user()->hasAccess('platform.systems.index'))
                 <a href="{{ route('platform.systems.index') }}" class="dropdown-item">
                     <x-orchid-icon path="settings" class="mr-2"/>
                     <span>{{ __('System') }}</span>
--- a/src/Access/UserSwitch.php
+++ b/src/Access/UserSwitch.php
@@ -19,10 +19,10 @@ class UserSwitch
     public static function loginAs(User $user)
     {
         if (! session()->has(self::SESSION_NAME)) {
-            session()->put(self::SESSION_NAME, Auth::id());
+            session()->put(self::SESSION_NAME, Auth::guard(config('platform.guard'))->id());
         }
 
-        Auth::login($user);
+        Auth::guard(config('platform.guard'))->login($user);
     }
 
     /**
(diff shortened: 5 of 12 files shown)


tabuna:
Hi, thanks for the work done. I want to clarify if it is impossible to specify a driver using middleware, for example:

```php
Route::middleware('auth:admin')->get('/user', function (Request $request) {
    return $request->user();
});
```

After all, we have the ability to specify our own rules in the configuration:

https://github.com/orchidsoftware/platform/blob/d0e5b15dce9dd688d51bf50019e0647acfc2a6d1/config/platform.php#L50-L53

ShamarKellman:
> Hi, thanks for the work done. I want to clarify if it is impossible to specify a driver using middleware, for example:
> 
> ```
> Route::middleware('auth:admin')->get('/user', function (Request $request) {
>     return $request->user();
> });
> ```
> 
> After all, we have the ability to specify our own rules in the configuration:
> 
> 
Thanks.

Yes you. can specify the middleware using 'auth:guard_name' . e.g. if you are using a guard called admin you can add `'auth:admin'` to the middleware like this `'private' => ['web', 'auth:admin', 'platform'], `

I will also add a pull request to the documentation with the usage if this pull request is accepted.

tabuna:
We have not come to an understanding. I mean, if we specify in the middleware:

```php
'private' => ['auth:admin'],
```

Then calls to `Auth::user()`/`Auth::check()` will already use the required guard without the need to specify explicitly: `Auth::guard(config('platform.guard'))`. Or I'm wrong?

ShamarKellman:
Based on the Laravel docs, when you are `Auth::user()` uses the default guard. When you are using multiple guards then you call `Auth::guard('custom_guard')->user()/Auth::guard('custom_guard')->check()`. 

The issue I was having was If I was using an admin guard with an Admin model with extends `Orchid\Models\User`, I would be able run `php artisan orchid:admin admin admin@admin.com password` which creates my user in the admins table. However when I go login I would be unable to login because the `Auth:check()` would be still checking the default guard which uses the App\Models\User which uses the users table. 

> 'private' => ['auth:admin'],
> Then calls to Auth::user()/Auth::check() will already use the required guard without the need to specify explicitly:  
> Auth::guard(config('platform.guard')). Or I'm wrong?

You will need to specify the guard explicitly so you would need to use `Auth::guard(config('platform.guard'))`.

tabuna:
Hi @ShamarKellman. I've spent some time researching how we can solve this. And I concluded that it would be enough to indicate the following:

1. We add a new option to the config:


```php
/*
|--------------------------------------------------------------------------
| Guard
|--------------------------------------------------------------------------
|
| This can be used if you are using a multi-auth setup configured.
| Such as using an Admin model for Orchid and User Model for frontend access.
| If not using default auth guard remember to add 'auth:guard_name' to the middleware
| where guard_name is the custom guard name.
|
| You can learn more here: https://laravel.com/docs/authentication
|
*/

'guard' => config('auth.defaults.guard', 'web'),
```

2. Then we change the entry point  `LoginController.php` (For jetstream/ui/etc you will need to do it yourself)

```diff
namespace Orchid\Platform\Http\Controllers;

use Illuminate\Auth\EloquentUserProvider;
use Illuminate\Contracts\Auth\Guard;
use Illuminate\Contracts\View\Factory;
use Illuminate\Cookie\CookieJar;
use Illuminate\Http\JsonResponse;
use Illuminate\Http\RedirectResponse;
use Illuminate\Http\Request;
- use Illuminate\Support\Facades\Auth;
+ use Illuminate\Contracts\Auth\Factory as Auth;
use Illuminate\Validation\ValidationException;
use Illuminate\View\View;
use Orchid\Access\UserSwitch;

class LoginController extends Controller
{
    /*
    |--------------------------------------------------------------------------
    | Login Controller
    |--------------------------------------------------------------------------
    |
    | This controller handles authenticating users for the application and
    | redirecting them to your home screen. The controller uses a trait
    | to conveniently provide its functionality to your applications.
    |
    */

+    /**
+     * @var Guard|\Illuminate\Auth\SessionGuard
+    */
+    protected $guard;

    /**
     * Create a new controller instance.
     *
     * @param Auth $auth
     */
-    public function __construct()
+    public function __construct(Auth $auth)
    {
+       $this->guard = $auth->guard(config('platform.guard'));

        $this->middleware('guest', [
            'except' => [
                'logout',
                'switchLogout',
            ],
        ]);
    }

    /**
     * Handle a login request to the application.
     *
     * @param Request $request
     *
     * @throws ValidationException
     *
     * @return JsonResponse|RedirectResponse
     */
    public function login(Request $request)
    {
        $request->validate([
            'email'    => 'required|string',
            'password' => 'required|string',
        ]);

-        $auth = Auth::guard()->attempt(
+        $auth = $this->guard->attempt(
            $request->only(['email', 'password']),
            $request->filled('remember')
        );

        if ($auth) {
            return $this->sendLoginResponse($request);
        }

        throw ValidationException::withMessages([
            'email' => __('The details you entered did not match our records. Please double-check and try again.'),
        ]);
    }

    /**
     * Send the response after the user was authenticated.
     *
     * @param \Illuminate\Http\Request $request
     *
     * @return \Illuminate\Http\RedirectResponse|\Illuminate\Http\JsonResponse
     */
    protected function sendLoginResponse(Request $request)
    {
        $request->session()->regenerate();

        return $request->wantsJson()
            ? new JsonResponse([], 204)
            : redirect()->intended(route(config('platform.index')));
    }

    /**
     * @param Request $request
+     * @param Guard   $guard
     *
     * @return Factory|View
     */
-    public function showLoginForm(Request $request, Guard $guard)
+    public function showLoginForm(Request $request)
    {
        $user = $request->cookie('lockUser');

        /** @var EloquentUserProvider $provider */
-        $provider = $guard->getProvider();
+        $provider = $this->guard->getProvider();

        $model = $provider->createModel()->find($user);

        return view('platform::auth.login', [
            'isLockUser' => optional($model)->exists ?? false,
            'lockUser'   => $model,
        ]);
    }

    /**
     * @param CookieJar $cookieJar
     *
     * @return RedirectResponse
     */
    public function resetCookieLockMe(CookieJar $cookieJar)
    {
        $lockUser = $cookieJar->forget('lockUser');

        return redirect()->route('platform.login')->withCookie($lockUser);
    }

    /**
     * @return RedirectResponse
     */
    public function switchLogout()
    {
        UserSwitch::logout();

        return redirect()->route(config('platform.index'));
    }

    /**
     * Log the user out of the application.
     *
     * @param \Illuminate\Http\Request $request
     *
     * @return \Illuminate\Http\RedirectResponse|\Illuminate\Http\JsonResponse
     */
    public function logout(Request $request)
    {
-        Auth::guard()->logout();
+        $this->guard->logout();

        $request->session()->invalidate();

        $request->session()->regenerateToken();

        return $request->wantsJson()
            ? new JsonResponse([], 204)
            : redirect('/');
    }
}
```

Thus, we will set our session to the desired `Guard`. After that, we only need to check in our middleware:

```diff
namespace Orchid\Platform\Http\Middleware;

use Carbon\Carbon;
use Closure;
use Illuminate\Contracts\Auth\Guard;
use Illuminate\Contracts\Routing\ResponseFactory;
use Illuminate\Http\RedirectResponse;
use Illuminate\Http\Request;
use Illuminate\Http\Response;
use Illuminate\Support\Facades\Route;
+ use Illuminate\Contracts\Auth\Factory as Auth;

/**
 * Class Access.
 */
class Access
{
    /**
     * @var Guard
     */
-    protected $auth;
+    protected $guard;

    /**
     * AccessMiddleware constructor.
     *
-     * @param Guard $auth
+     * @param Auth $auth
     */
-    public function __construct(Guard $auth)
+    public function __construct(Auth $auth)
    {
-        $this->auth = $auth;
+        $auth->shouldUse(config('platform.guard'));
+        $this->guard = $auth->guard(config('platform.guard'));

    }

    /**
     * @param Request $request
     * @param Closure $next
     * @param string  $permission
     *
     * @return ResponseFactory|RedirectResponse|Response|mixed
     */
    public function handle(Request $request, Closure $next, string $permission = 'platform.index')
    {
        Carbon::setLocale(config('app.locale'));

-        if ($this->auth->guest()) {
+        if ($this->guard->guest()) {
            return $this->redirectToLogin($request);
        }

-        if ($this->auth->user()->hasAccess($permission)) {
+        if ($this->guard->user()->hasAccess($permission)) {
            return $next($request);
        }

        // The current user is already signed in.
        // It means that he does not have the privileges to view.
        abort(403);
    }

    /**
     * Redirect on the application login form.
     *
     * @param Request $request
     *
     * @return \Illuminate\Contracts\Foundation\Application|ResponseFactory|RedirectResponse|Response
     */
    protected function redirectToLogin(Request $request)
    {
        if ($request->ajax() || $request->wantsJson()) {
            return response('Unauthorized.', 401);
        }

        if (Route::has('platform.login')) {
            return redirect()->guest(route('platform.login'));
        }

        if (Route::has('login')) {
            return redirect()->guest(route('login'));
        }

        abort(401);
    }
}
```

As a result, we set only on login, and then globally set in the middleware. As a result, we have avoided repeating the passing of the guard name every time we need to grab a user. That is, it will work correctly:

```php
Auth::user(); // return admin guard
````

What do you think about it?

ShamarKellman:
Much better than passing the config value. thanks.

tabuna:
[Released 10.0.0](https://github.com/orchidsoftware/platform/releases/tag/10.0.0)

PR #1626, opened 2021-03-19
Fix tabs not working on the same page if using same key

GuavaCZ:
Fixes that tabs are not switching correctly if multiple same keys on the same page.  

Fixed by adding a titles option, where you can enter the displayed name of the tabs. Keys should still be unique for each Tab-Layout.

Usage:
```php
Layout::tabs([
  'cz_1' => Layout::rows([
      // Your layout here
  ]),
  'de_1' => Layout::rows([
      // Your layout here
  ]),
], ["CZ", "DE"]), // This is what will be displayed
Layout::tabs([
    'cz_2' => Layout::rows([
        // Your layout here
    ]),
    'de_2' => Layout::rows([
        // Your layout here
    ]),
])->titles(["CZ", "DE]"), // This is what will be displayed
```

**Incorrect behavior:**
wrong

**Fixed behaviour:**
right

### DIFF
--- a/resources/views/layouts/tabs.blade.php
+++ b/resources/views/layouts/tabs.blade.php
@@ -12,7 +12,7 @@
                        id="button-tab-{{\Illuminate\Support\Str::slug($name)}}"
                        role="tab"
                        data-toggle="tab">
-                        {!! $name !!}
+                        {!! $titles[$loop->index] ?? $name !!}
                     </a>
                 </li>
             @endforeach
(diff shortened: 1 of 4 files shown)


tabuna:
Hi @GuavaCZ, what do you think if we just do independent tabs. 
Then we don't need the second argument and the `title` method?

GuavaCZ:
> 
> 
> Hi @GuavaCZ, what do you think if we just do independent tabs.
> Then we don't need the second argument and the `title` method?

Seems like a good idea, too. Do you mean something like multiple Layout::tab elements inside one parent Layout::tabs?

PR #1668, opened 2021-04-15
Fix menu sorting

plastique:
https://laravel.com/docs/8.x/collections#method-sort

> If your sorting needs are more advanced, you may pass a callback to `sort` with your own algorithm. Refer to the PHP documentation on `uasort`, which is what the collection's `sort` method calls utilizes internally.

### DIFF
--- a/src/Platform/Dashboard.php
+++ b/src/Platform/Dashboard.php
@@ -384,8 +384,8 @@ class Dashboard
     public function renderMenu(string $location): string
     {
         return $this->menu->get($location)
-            ->sort(function (Menu $menu) {
-                return $menu->get('sort', 0);
+            ->sort(function (Menu $current, Menu $next) {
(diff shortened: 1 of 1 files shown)


src/Platform/Dashboard.php
```diff
@@ -384,8 +384,8 @@ public function registerMenuElement(string $location, \Orchid\Screen\Actions\Men
     public function renderMenu(string $location): string
     {
         return $this->menu->get($location)
-            ->sort(function (Menu $menu) {
-                return $menu->get('sort', 0);
+            ->sort(function (Menu $menuA, Menu $menuB) {
```
tabuna, line 387: Hello, let's change the name of the variables to something more meaningful, for example:

```php
function (Menu $current, Menu $next)
```

plastique: Sorry, done.

PR #1675, opened 2021-04-18
Fix old name from request

mo7zayed:
Fixes #

Assume we have a input with name `article[tags][text][]` the `old()` input name should be `article.tags.text` in the current state it returns an extra `.` to the name `article.tags.text.`

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -49,7 +49,6 @@
     <env name="SCOUT_DRIVER" value="array"/>
     <env name="SCOUT_QUEUE" value="true"/>
     <env name="QUEUE_DRIVER" value="sync"/>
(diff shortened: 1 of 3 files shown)


tabuna:
Hi @mo7zayed, can we add a test?

mo7zayed:
@tabuna Done. Check please :smile:

tabuna:
Let's add a check `withErrors(...)`. So that we know for sure that the error message is displayed

tabuna:
Hi @mo7zayed  do you need help?

mo7zayed:
@tabuna
Thanks :smile: , I was super busy yesterday, I'm working on it now

mo7zayed:
@tabuna Check now

PR #1693, opened 2021-04-25
adding title property to Table

danieljonguitud:
## Proposed Changes

  - Add optional title to tables

### DIFF
--- a/resources/views/layouts/table.blade.php
+++ b/resources/views/layouts/table.blade.php
@@ -1,3 +1,8 @@
+@if($title)
+    <div class="px-3 mb-4 mt-4">
+        <h4>{{$title}}</h4>
+    </div>
+@endif
 <div class="bg-white rounded shadow-sm mb-3"
      data-controller="table"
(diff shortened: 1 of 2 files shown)


tabuna:
Hi @danieljonguitud. I understand why you made a separate method and property. But we must be consistent in each object so as not to irritate. We already have classes where the title is declared differently and supports short syntax.

```php
Layout::rows([...])->title('Header')
Layout::table('target', [...])->title('Header')
```

I think we should change this and not change, at least without breaking BC. Here's a small example where you can take a look:

https://github.com/orchidsoftware/platform/blob/91b0a3361c8bae84236d756fd673e6e3a51fb52e/src/Screen/Layouts/Rows.php#L64-L69

I also have a question about appearance. In the previous variation, we use the `<legend>` tag. Maybe we could also use it too?

danieljonguitud:
Thank you, @tabuna, for your feedback. I did the changes you suggested and I will try to be more cautious about the project consistency. If there else something else to add or any testing to be made please let me know.

PR #1745, opened 2021-05-25
Added config for fallback routes setting

falur:
Fixes # if DASHBOARD_DOMAIN == / api routes did not work

## Proposed Changes

  - Added config for fallback routes setting
  - Added (if) to dashboard file with routes


config/platform.php
```diff
@@ -193,4 +193,19 @@
         'cache' => false
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Routes
```
tabuna, line 198: Let's change the title to "Fallback Page".

config/platform.php
```diff
@@ -193,4 +193,19 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | Default configuration for platform routes
+    |
+    */
+
+    'routes' => [
```
tabuna, line 205: If we make a group, then it is strange that there is only one element in it.
Let's not use a multidimensional array for now.

```php
'fallback'  => true,
```

config/platform.php
```diff
@@ -193,4 +193,19 @@
         'cache' => false
     ],
 
+    /*
+    |--------------------------------------------------------------------------
+    | Routes
+    |--------------------------------------------------------------------------
+    |
+    | Default configuration for platform routes
```
tabuna, line 201: What do you think about adding such a description?

> If the request does not match any route and arguments, Orchid will automatically generate its own 404 page. It can be disabled if you want to declare routes on the same domain and prefix or create your own page.

config/platform.php
falur: А после может и префикс с доменом в одну секцию routes поместить, в мажорном обновлении. Это так полёт фантазии

config/platform.php
falur: Super!

PR #1782, opened 2021-06-24
Added new events for uploading files

falur:
## Proposed Changes
Added new events for uploading files
  - ReplicateFileEvent - for replicate attachment
  - UploadedFileEvent - for all uploading files


src/Platform/Http/Controllers/AttachmentController.php
```diff
@@ -111,6 +112,8 @@ private function createModel(UploadedFile $file, Request $request)
 
         $model->url = $model->url();
 
+        event(new UploadedFileEvent($model));
```
tabuna, line 115: Wouldn't that cause the event to be discarded twice, as it is in the File class called above?

falur: it's different events UploadFileEvent and Upload**ed**FileEvent

PR #1802, opened 2021-07-16
Add Catalan translation

iroes:


iroes:
Thanks especially for the promptness!

PR #1814, opened 2021-07-31
add custom title for screen

longlp:
Fixes #

## Proposed Changes

  - Show custom title for screen when needed.
  -
  -


tabuna:
Hi, could you please describe the practical benefits of this change

longlp:
It is the screen_title, if not define the layout will show title.
use screen_title when we want to show the orthers format such as link, or a html code.
For example: I did a File Explorer, the the title is the path of the current folder, but i want to show the path as Link (/root/abc/def)
- root - link to root folder
- abc - link to root/abc and so on.
the existed version use only title for all titles.

tabuna:
To do this, you can use the bread kroshli

PR #1853, opened 2021-08-29
Added pt_BR (Brazilian Portuguese) translation

wisliy:
Fixes #

## Proposed Changes
Added pt_BR (Brazilian Portuguese) translation


tabuna:
Hi 👋 Thank you very much for the work done, but there are some lines in the translations that are not used in this package.  Can you pay attention to passing tests?

wisliy:
@tabuna, I added some strings from User, Roles and Profile pages that were not translated (lines 109 to 133 and seems like they are not passing the tests. We can't add this missing strings?

wisliy:
@tabuna, now I added with the same strings ru.json file. Hope it works.

PR #1858, opened 2021-09-02
Added routeprefixed class to body

SimonErich:
Added a class to body, that will always use a slugified version of the current path as basic.
This allows to style subpages (login, template, ...) differently easily using a custom css file.

### DIFF
--- a/resources/views/app.blade.php
+++ b/resources/views/app.blade.php
@@ -43,7 +43,7 @@
     @endforeach
 </head>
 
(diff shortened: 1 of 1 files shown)


tabuna:
Hi @SimonErich, maybe we'd better use the route name?
After all, if there is a parameter, for example `/user/1/posts/hello-world`, then it will no longer work just

SimonErich:
thanks @tabuna . Good idea. I have updated it.

PR #1863, opened 2021-09-09
select fix on int key

mucan54:
Int and string number pair fix.


tabuna:
Hey @mucan54, thanks for your interest and work. 
I see that automated tests are failing. Can you figure it out, as well as write your check?

PR #1911, opened 2021-10-08
Improve TEMPLATE_LINE filter

grigoriy-ivanov:
Added margin for button group. 
Added text for buttons for better  UX

Before:
изображение

After:
изображение


grigoriy-ivanov:
Maybe its better.
I used css from Screen

PR #1927, opened 2021-10-17
support for complex TD::filters, like daterange, number interval

lintaba:
Fixes #

## Proposed Changes

  - new field type: numberRange
  - HttpFilter supports array typed queries
  - HttpFilter can now filter both date ranges, and number ranges. (date ranges needs model to cast the field to date-like)
  - TD, new filter types: FILTER_DATE_RANGE, FILTER_NUMBER_RANGE, FILTER_SELECT
  - new method: TD@filterOptions, to set up possible options for select-filter
  - TD@filter, new (optional) second parameter, to initialize select-filter. (shorthand for TD@filterOptions)
  - TD@sluggable is now protected instead of private (supports better customization)
  - helpers: HttpFilter is resolved via service container instead of hard-coded class, for better customization

image


lintaba:
hey, you are right with the select, but i think the range filters are common enough to be there as constants, and gives better explorability, in case the documentation lags behind.

will check the tests as i have some time for it

AidasK:
Any update? Seems more than useful

src/Filters/HttpFilter.php
```diff
@@ -66,7 +66,9 @@ protected function parseHttpValue($query)
         if ($query === null) {
             return null;
         }
-
+    
+        if(is_array($query)){return $query;}
```
AidasK, line 70: psr-2 formatting

src/Filters/HttpFilter.php
```diff
@@ -132,15 +134,25 @@ protected function addFiltersToQuery(Builder $builder)
     protected function filtersExact(Builder $query, $value, string $property): Builder
     {
         $property = self::sanitize($property);
-
+    
+        if ($query->getModel()->hasCast($property, ['date', 'datetime', 'immutable_date', 'immutable_datetime']) || $property === $query->getModel()->getCreatedAtColumn() || $property === $query->getModel()->getUpdatedAtColumn()) {
```
AidasK, line 138: what about deleted_at? maybe these columns already has casts on them and we don't need to check the name explicitly?

src/Filters/HttpFilter.php
```diff
@@ -132,15 +134,25 @@ protected function addFiltersToQuery(Builder $builder)
...
+            if (is_array($value) && (isset($value['start']) || isset($value['end']))) {
+                return $query->whereDate($property, '<=', $value['end'] ?? now()->addYears(50))
```
AidasK, line 140: can be changed to for more optimal queries:
```php
$query->when($value['start'] ?? null, fn (Builder $query) => $query->whereDate($property, '>=', $value['start']));
$query->when($value['end'] ?? null, fn (Builder $query) => $query->whereDate($property, '<', $value['end']));
```

src/Filters/HttpFilter.php
```diff
@@ -132,15 +134,25 @@ protected function addFiltersToQuery(Builder $builder)
...
+                             ->whereDate($property, '>=', $value['start'] ?? now()->subYears(50));
+            }
+        }
+        if (is_array($value) && (isset($value['start']) || isset($value['end']))) {
```
AidasK, line 144: number range should be called min,max and should be used similar approach as above

src/Screen/TD.php
```diff
@@ -300,21 +327,21 @@ public function cantHide($hidden = false): self
...
     }
-
+    
     /**
      * @return string
      */
     public function buildSortUrl(): string
     {
-        $query = request()->query();
+        $query         = request()->query();
```
AidasK, line 339: strange formatting

src/Support/helpers.php
```diff
@@ -73,7 +73,7 @@ function get_sort(?string $property): string
      */
     function get_filter(string $property)
     {
-        $filter = new HttpFilter();
+        $filter = app(HttpFilter::class);
```
AidasK, line 76: why do we need DI inection for http filter?

src/Filters/HttpFilter.php
lintaba: no need, its a datetime cast by default:
```
# SoftDeletes.php
public function initializeSoftDeletes()
    {
        if (! isset($this->casts[$this->getDeletedAtColumn()])) {
            $this->casts[$this->getDeletedAtColumn()] = 'datetime';
        }
    }
```

src/Support/helpers.php
lintaba: So it can be overriden when needed. eg. if an user wants to be able to expand filtersExact further (as I needed with arrays), they can do it. Without this extension, it needs a lots of hack. With adding DI they can just add a DI override within a service provider.

AidasK: not sure it's going to be approved like this. Please change it back as it's not a part of this mr

src/Filters/HttpFilter.php
```diff
@@ -134,16 +136,19 @@ protected function addFiltersToQuery(Builder $builder)
...
-        if ($query->getModel()->hasCast($property, ['date', 'datetime', 'immutable_date', 'immutable_datetime']) || $property === $query->getModel()->getCreatedAtColumn() || $property === $query->getModel()->getUpdatedAtColumn()) {
-            if (is_array($value) && (isset($value['start']) || isset($value['end']))) {
-                return $query->whereDate($property, '<=', $value['end'] ?? now()->addYears(50))
-                             ->whereDate($property, '>=', $value['start'] ?? now()->subYears(50));
-            }
+        
+        if ($query->getModel()->hasCast($property, ['date', 'datetime', 'immutable_date', 'immutable_datetime']
+            ) || $property === $query->getModel()->getCreatedAtColumn() || $property === $query->getModel()
+                                                                                               ->getUpdatedAtColumn()) {
+            $query->when($value['min'] ?? null, fn (Builder $query) => $query->whereDate($property, '>=', $value['min']));
```
AidasK, line 143: some miscommunication has happened. min max should only be used for numbers and start end for dates as it sounds more consistent with what we are checking.

AidasK:
Tests are failing and tests are needed for number range

src/Filters/HttpFilter.php
lintaba: I'd use the same terminology for both. So either min/max, or start/end in both places.

lintaba: okay, now I see. Will revert the changes within daterange, and keep it only in numberrange.

lintaba:
> Tests are failing and tests are needed for number range

should be all good now.

lintaba:
> Tests are failing and tests are needed for number range

@AidasK, should be all good now.

AidasK:
Have you also added some new tests for new data fields?

AidasK:
@tabuna, @swara-mohammed can you review this MR? @lintaba is really trying to finish up this feature.

PR #1953, opened 2021-11-08
Allow show/hide columns for unpaginated tables #fix

iroes:
Fix for displaying the 'paginator' footer when there are columns to show/hide, in case the queried Collection is not paginated.


resources/views/layouts/table.blade.php
```diff
@@ -63,7 +63,10 @@
         </div>
     @endif
 
-    @includeWhen(($rows instanceof \Illuminate\Contracts\Pagination\Paginator || $rows instanceof \Illuminate\Contracts\Pagination\CursorPaginator) && $rows->isNotEmpty(),
+    @includeWhen(
```
tabuna, line 66: I think the condition is already very long. Let's try to remove it and indicate the missing part in `platform:: layouts.pagination`

iroes: I think I don't get it. Why not just remove the condition? Isn't already indicated?

PR #1955, opened 2021-11-09
Add close icon for selected filters

grigoriy-ivanov:
Before
изображение

After
изображение


tabuna:
Hey @grigoriy-ivanov. Don't your users understand these badges? Or do you think it would be better this way?

grigoriy-ivanov:
Yes, they don't. 
I think it looks better this way, like a multiple select field with selected options.
Examples:
Ozon:
изображение
DNS:
изображение

tabuna:
Ping @Hesko ^

tabuna:
@grigoriy-ivanov Thanks a lot. We changed it a bit and made a release.

PR #1971, opened 2021-11-20
Accepts the file path as an optional argument

mahmoudmohamedramadan:
This PR for allowing to pass the file path as an optional argument because it's NOT nice to force me to save the file always in a `Date` folder now after this PR you can pass the folder path that you want to save the file in

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 10.4.2 - 2021-05-21
+
+### Fixed
+- Added padding for date-filter
+- Display divider for empty profile menu
+
 ## 10.4.1 - 2021-05-21
 
 ### Fixed
--- a/src/Platform/Dashboard.php
+++ b/src/Platform/Dashboard.php
@@ -18,7 +18,7 @@ class Dashboard
     /**
      * ORCHID Version.
      */
-    public const VERSION = '10.4.1';
+    public const VERSION = '10.4.2';
 
     /**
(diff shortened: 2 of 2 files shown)


src/Screen/Fields/Relation.php
```diff
@@ -92,7 +92,12 @@ class Relation extends Field
      */
     public function fromModel(string $model, string $name, string $key = null): self
```
mahmoudmohamedramadan, line 93: This is old PR and I forgot to pull BEFORE pushing a new feature 😅

tabuna: Please do a rebase to avoid this.

src/Attachment/File.php
```diff
@@ -77,7 +84,9 @@ public function load(): Model
     {
         $attachment = $this->getMatchesHash();
 
-        if (! $this->storage->has($this->engine->path())) {
+        if (!is_null($this->path)) {
```
tabuna, line 87: I don't think testing in this way is a good solution. It would be much better if we changed this to allow passing a parameter to the engine class. An example that fits here and below:
```diff
- $this->path ?? $this->engine->path()
+ $this->engine->path($path)
```

And the engine class itself decided how to process this information—returning the same path.

src/Attachment/File.php
```diff
@@ -52,14 +57,16 @@ class File
      *
      * @param UploadedFile $file
      * @param string|null  $disk
+     * @param string|null  path
      * @param string|null  $group
      */
-    public function __construct(UploadedFile $file, string $disk = null, string $group = null)
+    public function __construct(UploadedFile $file, string $disk = null, string $path = null, string $group = null)
```
tabuna, line 63: Warning: This is a breaking change and therefore cannot be released under the 10.x branch

mahmoudmohamedramadan:
Hi sir,

Can you please take a look over my new PR here
https://github.com/orchidsoftware/platform/pull/1973

Issue #969, opened 2019-09-26
Make class function of input field add to existing classes


2kodes:
**Is your feature request related to a problem? Please describe.**
I added a class `mt` to a checkbox like this 

`CheckBox::make('field')->title('Allow download?')->class('mt-4')`

When the field was rendered only the class I added was in the class attribute. I think the class I added should have been added to the already existing classes for the field.

**Describe the solution you'd like**
1. The new class is appended to the existing class for the field or html element when rendered 
      OR
2. The new class is appended by default but there is an option to specify if it should override the existing class

tabuna:
I don't think adding a default would be a good solution. This causes a mismatch with other methods.

For example:

```php
CheckBox::make('field')
    ->value('Default value')
    ->value('Custom value 2');
```
I expect to receive `Custom value 2`, but not their gluing.

In this case, it would be great to have additional methods like:

```php
CheckBox::make('field')
    ->class('form-control mt-1')
    ->addClass('mt-4')
    ->removeClass('mt-1');
```

As a result of this: `form-control mt-4`

tabuna:
There may be several options for implementation.

- Create and use a trade
- Create a basic concatenation method and use trade / methods
- Or make them magical, by attribute name

2kodes:
I like `addClass`  since the results will be clear . In this case we can still keep the `class` method for cases when you want to override

2kodes:
> There may be several options for implementation.
> 
> * Create and use a trade
> * Create a basic concatenation method and use trade / methods
> * Or make them magical, by attribute name

I am not exactly clear on this

2kodes:
Hello @tabuna , 

Please any update on this??

tabuna:
Hi @2kodes .

This is not a mistake, and accordingly it has a rather low priority. 
I work in my free time, so this may not happen quickly. 
But you can help me by making PR.

Even if you cannot and do not want to do this, we will add it anyway, but you will have to wait;)

2kodes:
lol.. cool .. no problem.will see if i can make a PR

lhoucinecherif:
Hi,

I made a Workaround for this , you can add your own functions to a Trait like this : 

create an extended class for every Field you want to use  (here is an example with a DateTimer ) : `./app/Orchid/Fields/DateTimer.php`

This class will extend the original Field,  and use our custom Trait  like this : 

```
<?php

namespace  App\Orchid\Fields;

use \App\Orchid\Traits\Field;

[...]

PR #1972, opened 2021-11-20
add hasClass, addClass and removeClass

lhoucinecherif:
Adding hasClass, addClass and removeClass methods to the fields

Fixes #

Make class function of input field add to existing classes #969
https://github.com/orchidsoftware/platform/issues/969


tabuna:
Hi, can you add tests?

PR #1988, opened 2021-12-01
BugFix Trait AsSource

ivanomatteo:
The trait is part of the model class, so can access also to private and protected members.

For example if we use this trait on a model with a nullable database field called "visible" 

Arr::get($this->toArray(), $field) ---> is null
Arr::get($this->getRelations(), $field) ---> is null
$this->$field ---> refear to "protected $visible = []"

with this commit the problem is solved, and anso bring a performance improvement:
->toArray() force the execution of all casts and accessors, but only one filed is requested.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/Screen/AsSource.php
+++ b/src/Screen/AsSource.php
@@ -18,8 +18,9 @@ trait AsSource
      */
     public function getContent(string $field)
     {
-        return Arr::get($this->toArray(), $field)
-            ?? Arr::get($this->getRelations(), $field)
-            ?? $this->$field;
+        if ($this->isRelation($field) && !$this->relationLoaded($field)) {
+            return null;
+        }
+        return $this->getAttribute($field);
     }
 }


src/Screen/AsSource.php
```diff
@@ -18,8 +18,9 @@ trait AsSource
      */
     public function getContent(string $field)
     {
-        return Arr::get($this->toArray(), $field)
-            ?? Arr::get($this->getRelations(), $field)
-            ?? $this->$field;
+        if ($this->isRelation($field) && !$this->relationLoaded($field)) {
```
tabuna, line 21: These lines will break the usual behavior that users are accustomed to. For example with access to private properties to be prohibited, it was enough to change only
 the call to `$this->$field` to` $this->getAttribute($field) `

ivanomatteo: Ok, do you mean when passing $field === null?

in that case just prepend:

if($field === null){
return $this->toArray();
}

tabuna: I mean, for solution #1989 it is enough to change the call to `getAttribute`.
Other changes do not affect issue #1989. Plus, they break current tests (and existing user code.)

https://github.com/orchidsoftware/platform/blob/39d631aa2c09c247bfe45d802ff2ff81448ef15b/tests/Unit/Screen/SourceTest.php#L67-L91

ivanomatteo: Oh, I see, it must support dot notation.

PR #2023, opened 2021-12-26
add accepted-files to cropper and picture

G3z:
Adds the accepted method to both picture and cropper to control uploadable mime

### DIFF
--- a/src/Screen/Fields/Cropper.php
+++ b/src/Screen/Fields/Cropper.php
@@ -8,7 +8,6 @@ namespace Orchid\Screen\Fields;
  * Class Cropper.
  *
  * @method Cropper accept($value = true)
- * @method Picture acceptedFiles(string $value = null)
  * @method Cropper accesskey($value = true)
  * @method Cropper autofocus($value = true)
  * @method Cropper disabled($value = true)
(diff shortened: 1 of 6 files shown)


src/Screen/Fields/Cropper.php
```diff
@@ -65,7 +67,8 @@ class Cropper extends Picture
      * @var array
      */
     protected $inlineAttributes = [
-        'accept',
+        '
+        @method Picture acceptedFiles(string $value = null)
```
tabuna, line 71: Check it out, I think there is a typo here

src/Screen/Fields/Cropper.php
```diff
@@ -57,6 +58,7 @@ class Cropper extends Picture
         'maxHeight'      => 'Infinity',
         'maxFileSize'    => null,
         'staticBackdrop' => false,
+        'acceptedFiles'  => null,
```
tabuna, line 60: Maybe set the default value `image/*` so that only images are in the review?

src/Screen/Fields/Cropper.php
G3z: I've fixed this

PR #2024, opened 2021-12-26
Variable for vertical orientation in Block class

astersnake:
## Proposed Changes

  - `vertical` variable and function add to block class 
  - if `vertical` is true in `block.blade.php` layout orientation will be vertical instead of horizontal
  
The reason behind this is to have the ability to use full horizontal space for layouts that have more information (tables for example) without missing the functionality of the Block class.


src/Screen/Layouts/Block.php
```diff
@@ -37,6 +37,7 @@ abstract class Block extends Layout
     public function __construct(array $layouts = [])
     {
         $this->layouts = $layouts;
+        $this->variables['vertical'] = false;
```
tabuna, line 40: Let's use the property:

```php
protected $layouts = [
    'vertical' => false
];
```

resources/views/layouts/block.blade.php
```diff
@@ -8,7 +8,7 @@
             </p>
         </legend>
     </div>
-    <div class="col-12 col-md-7 shadow-sm">
+    <div class="col-12 {{$vertical ? 'col-md-12' : 'col-md-7'}} shadow-sm">
```
tabuna, line 11: By default, full-width markup is used. We don't need to re-install it for a specific screen.

src/Screen/Layouts/Block.php
```diff
@@ -86,6 +87,20 @@ public function description($description): self
...
     }
+    
+    /**
+     * Used to define block orientation.
+     *
+     * @param bool $vertical
+     *
+     * @return Block
+     */
+    public function vertical($vertical): self
```
tabuna, line 98: Let's make it possible to call this method without constantly passing the value `true`

```php
public function vertical($vertical = true)
```

astersnake: Ok, first I thought this but codeclimate scared me lol

src/Screen/Layouts/Block.php
astersnake: I tried this but I got `Undefined variable`

astersnake: Nevermind, the variable name was `$variables`

resources/views/layouts/block.blade.php
astersnake: So this should be updated to 

`` <div class="col-12 {{!$vertical ? 'col-md-7' : ''}} shadow-sm">``

Right?

astersnake:
@tabuna I updated the pull request according to your suggestions

PR #2042, opened 2021-12-29
Remove buildDd return type in Sight 

astersnake:
## Proposed Changes

  - Make Sight `buildDd()` function have the same return type (none) as `buildDd()` in `TD` so the render function can have the same behavior.

### DIFF
--- a/src/Screen/Sight.php
+++ b/src/Screen/Sight.php
@@ -31,7 +31,7 @@ class Sight extends Cell
      *
      * @return string|null
      */
-    public function buildDd($repository): ?string
+    public function buildDd($repository)
     {
         $value = $this->render
             ? $this->handler($repository)


tabuna:
Hi, can you give an example where this behavior gets in the way?

hm-alishah:
@tabuna here is an example of what I am trying to accomplish but the forced return gets in the way
```
   Sight::make('Status')->render(fn() => $this->canceled ? Color::DANGER(): Color::INFO()),
```

tabuna:
Hey @hm-alishah 

For your objects to be returned, they must have [tostring](https://www.php.net/manual/en/language.oop5.magic.php#language.oop5.magic.tostring). 
For PHP 7.4 and below, you must explicitly indicate that you want the string. Example:

```php
(string) $object;
```

PR #2064, opened 2022-01-15
Adding compact option to Table layout

mvehar-preskok:
## Proposed Changes

Adding compact Table layout (compact bootstrap table)

### DIFF
--- a/resources/sass/core/tables.scss
+++ b/resources/sass/core/tables.scss
@@ -1,3 +1,5 @@
+@import '~bootstrap/scss/variables';
+
 .table {
   thead {
     tr {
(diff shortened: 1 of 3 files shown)


tabuna:
Have you checked it? For me, adding the class didn't change anything. Probably CSS was not overridden

mvehar-preskok:
There was an issue with ccs rules priority. Decided to add new class that copies the original table-sm + adding important. 
I could also use exact css rule as the one that originally prevented the table-sm to work.

P.s. Should I include all built files in PR? vendor.js, orchid.js, orchid.css, etc.?

tabuna:
Please don't include js/css build files in your PR

resources/sass/core/tables.scss
```diff
@@ -1,3 +1,5 @@
+@import '~bootstrap/scss/variables';
```
tabuna, line 1: Variables are already loaded earlier. You don't have to re-upload them.

PR #2069, opened 2022-01-17
Solve Radio buttons id duplicated

swara-mohammed:
Fixes #

## Proposed Changes

  - Unique id for radio button


tabuna:
Can you give an example of what this change fixes before and after?

swara-mohammed:
In the pictures below, it may be clear what I mean. IDs are repeated because each component(field) has one unique id.
Bildschirmfoto 2022-01-20 um 11 32 16

Bildschirmfoto 2022-01-20 um 11 32 49

swara-mohammed:
With the changes I've made, each id will first get its own key[from options] and then the field's id. That's how it becomes a unique id.

PR #2071, opened 2022-01-18
Allow to load more attachments in upload existing media

thewebartisan7:
At the moment you can only search attachments by original name which can be hard to find what you are looking for.

## Proposed Changes

  - Added load more button in upload.blade.php view modal
  - Added loadMore() method in upload_controller.js
  - Added pagination result in AttachmentController.php

## Behaviour

If you click on load more button, then results it's appended.
If you search via input, then results it's replaced.
If there are not anymore page to load, load more button is hidden.

## Note 

I was not sure how to handle append or replace, so I use the event object passed by input search, and when it's undefined, that means load more button is clicked. Since I have not yet enough experience with hotwired, maybe there is a better way to handle this.

## Demo

[Image from Gyazo](https://gyazo.com/040242c34589ff712287646d6bbac644)


thewebartisan7:
Updated file  resources/views/fields/upload.blade.php  for making modal scrollable.

tabuna:
https://github.com/orchidsoftware/platform/releases/tag/10.31.0

PR #2072, opened 2022-01-18
Add a custom event orchid:listener:post-render

G3z:
Add a custom event triggered when listener has finished updating the elements

Fixes it is possible to run js after a lister has updated the elements, for example it's possible to rehydrate Livewire components with

```js
document.addEventListener("orchid:listener:post-render", () => {
    Livewire.start();
});
```

### DIFF
--- a/resources/js/controllers/listener_controller.js
+++ b/resources/js/controllers/listener_controller.js
@@ -46,7 +46,7 @@ export default class extends ApplicationController {
 
         this.asyncLoadData(params).then(() => {
             document.dispatchEvent(
-                new CustomEvent("orchid:listener:post-render", {
+                new CustomEvent("orchid:listener:after-render", {
                     detail: {
                         params: params,
                     },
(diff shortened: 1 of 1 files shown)


resources/js/controllers/listener_controller.js
```diff
@@ -44,7 +44,15 @@ export default class extends ApplicationController {
                 }
             }));
 
-        this.asyncLoadData(params);
+        this.asyncLoadData(params).then(() => {
+            document.dispatchEvent(
+                new CustomEvent("orchid:listener:post-render", {
```
tabuna, line 49: What do you think about renaming it to `orchid:listener:after-render for more clarity?

tabuna:
I haven't worked much with livewire and they have a separate method for reloading:

```js
document.addEventListener("orchid:listener:after-render", () => {
    window.Livewire.restart();
});
```

resources/js/controllers/listener_controller.js
G3z: sure!

G3z:
just for info, when do you plan to release this ?

tabuna:
Hey @G3z just https://github.com/orchidsoftware/platform/releases/tag/10.30.0

PR #2080, opened 2022-01-24
Destroy popover on disconnect

thewebartisan7:
Fixes 

Destroy popover on disconnect

## Proposed Changes

  - Update popover_controller.js

### DIFF
--- a/resources/js/controllers/popover_controller.js
+++ b/resources/js/controllers/popover_controller.js
@@ -19,4 +19,11 @@ export default class extends ApplicationController {
         event.preventDefault();
         this.popover.toggle();
     }
+    
+    /**
+     * Destroy popover on disconnect
+     */
+    disconnect() {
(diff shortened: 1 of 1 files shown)


thewebartisan7:
I think would be also good to not initialize popover if page is cached.

Below is full proposed change:

```js
import ApplicationController from "./application_controller";
import { Popover } from 'bootstrap';

export default class extends ApplicationController {
    /**
     *
     */
    connect() {
        if (document.documentElement.hasAttribute("data-turbo-preview")) return
        this.popover = new Popover(this.element, {});
    }

    /**
     *
     * @param event
     */
    toggle(event) {
        event.preventDefault();
        this.popover.toggle();
    }

    /**
     *
     */
    disconnect() {
        if (!this.popover) return
        this.popover.dispose();
    }
}
```

PR #2082, opened 2022-01-25
Screen reader classes are now visually-hidden

thewebartisan7:
Renamed .sr-only and .sr-only-focusable to .visually-hidden and .visually-hidden-focusable

See https://getbootstrap.com/docs/5.1/migration/#helpers

### DIFF
--- a/resources/views/fields/upload.blade.php
+++ b/resources/views/fields/upload.blade.php
@@ -146,7 +146,7 @@
                                         </div>
 
                                         <div class="media-loader spinner-border" role="status">
-                                            <span class="sr-only">{{ __('Loading...') }}</span>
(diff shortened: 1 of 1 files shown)


thewebartisan7:
Suggestion also to center the loader:

Update HTML structure like below:

```html
                                        <div class="media-loader">
                                            <div class="spinner-border" role="status">
                                                <span class="visually-hidden">{{ __('Loading...') }}</span>
                                            </div>
                                        </div>
```

Add CSS:

```css
.media-loader {
    display: flex;
    justify-content: center;
}
```

We can't use utility classes .d-flex and .justify-content-center because inline style for hide media-loader will not works.

Result:

Screenshot 2022-01-25 at 04 51 33

PR #2101, opened 2022-02-01
Add variable to make remove rows optional in matrix field

astersnake:
## Proposed Changes

  - Sometimes matrix could be use to update data but not remove it, so should it exist the option to disable the default trash button in every row.

### DIFF
--- a/src/Screen/Fields/Matrix.php
+++ b/src/Screen/Fields/Matrix.php
@@ -96,7 +96,7 @@ class Matrix extends Field
      *
      * @return Field|Matrix
      */
-    public function removableRows(bool $value)
+    public function removableRows(bool $value = true)
     {
         return $this->set('removableRows', $value);
     }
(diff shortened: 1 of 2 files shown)


src/Screen/Fields/Matrix.php
```diff
@@ -90,6 +91,16 @@ public function maxRows(int $count)
         return $this->set('maxRows', $count);
     }
 
+    /**
+     * @param bool $value
+     *
+     * @return Field|Matrix
+     */
+    public function removableRows(bool $value)
```
tabuna, line 99: I think that constantly passing a value to a method will not be very convenient. 
Why don't we set the default value to true, just like we do in other Field methods?

astersnake: Yes, I forgot the default value in the function. I just did a commit with that.

astersnake:
@tabuna it's possible to accept this pull request?

PR #2108, opened 2022-02-03
Fix profile `view` with a long name / role

### LINKED ISSUE #2107: The layout breaks when the role name is long
When using long role names or when using multiple roles for one user, the user's avatar size decreases.

This is true for the latest version of the platform 10.31.0.

drabodan:
Fix #2107

## Proposed Changes

  - profile show issues if name and/or roles are too long

Before:

Screenshot_2

After:

Screenshot_3

### DIFF
--- a/resources/views/partials/profile.blade.php
+++ b/resources/views/partials/profile.blade.php
@@ -1,12 +1,12 @@
 <div class="d-flex m-0 align-items-center p-3">
-    <div class="dropdown w-100">
+    <div class="dropdown w-75">
         <a href="#" class="nav-link p-0 d-flex align-items-center" data-bs-toggle="dropdown">
             @if($image = Auth::user()->presenter()->image())
                 <span class="thumb-sm avatar me-3">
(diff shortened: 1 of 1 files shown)


tabuna:
It will move the notification icon. 

image

The Role enumeration is a simple example. 
There is no obligation to use it. It can be a position, a balance sheet, or even an email.

We can cut this down at the presentors class level:

```php
    public function subTitle(): string
    {
        $roles = $this->entity->roles->pluck('name')->implode(' / ');
        
        return (string) \Illuminate\Support\Str::of($roles)
            ->limit(20)
            ->whenEmpty(function () {
                return __('Regular user');
            });
    }
```

drabodan:
Ah...it was stupid, I completely forgot about the notification (I checked it with it disabled). Thanks for your review!

Perhaps then, in order not to complicate the logic, we can use inline styles, as we already did for the `name`, only reduce the value to 9.5 so that the text fits and apply this to the entire text block. What do you think?

PR #2110, opened 2022-02-05
Update modal_controller

Augen2203:
Fixes :

1. When we submit a form and it returns with a validation error, its title was reset.

2. When we use asynchronous data loading in the modal window and the screen is updated, the asynchronous data was not requested again.

3. If we used several modals on the same screen, and at the same time they had the same form layouts, when we receive a validation error in one of them, we will be shown all of them (since they all fit the old condition).

> As an example:
> 
> **One screen** user list
> 
> and two modals with the same field layouts
> 
> `Layout::modal('userEditAsyncModal', UserEditLayout::class)->async('asyncGetUser')`
>             
> `Layout::modal('userCreateAsyncModal', UserEditLayout::class)`
> 
> In case of a validation error in one of them, they will cause the bug mentioned above
> 
> The fix is ​​achieved by checking the slug of the last opened modal window, and opens it in case of an error.
> 

4. Partial ability to use tables in a modal window, since now when paginating or using a filter, the modal window reopens, and does not remain closed as it was before.

**Briefly about the change in the behavior scheme:**

1. In `sessionStorage`, we now store the json data string of the modal window, which allows us to save: the title, the form action, the parameters for the asynchronous request

> Recording example:
> `{"title":"Edit user","submit":"http://example.com/users/saveUser?user=5","params":"{\"user\":5}","slug ":"aad113213c0e17a5038d2b6cea7f16f87c7ae37d","validateError":true}`

2. If the closing event of the modal window fires, its `sessionStorage` is completely cleared

### DIFF
--- a/resources/js/controllers/modal_controller.js
+++ b/resources/js/controllers/modal_controller.js
@@ -10,43 +10,48 @@ export default class extends ApplicationController {
         "title"
     ];
 
+    initialize() {
+        this.show = this.show.bind(this);
+        this.hidden = this.hidden.bind(this);
+    }
+
     /**
      *
      */
     connect() {
+        this.element.addEventListener('shown.bs.modal', this.show);
+        this.element.addEventListener('hide.bs.modal', this.hidden);
+        this.openLastModal();
+    }
 
-        $(this.element).on('shown.bs.modal', () => {
-            let autoFocusElement = this.element.querySelector('[autofocus]');
+    /**
+     * Show EventListener
+     */
+    show(event)
+    {
+        let autoFocusElement = this.element.querySelector('[autofocus]');
 
-            if(autoFocusElement !== null){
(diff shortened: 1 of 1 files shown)


Augen2203:
Thanks for fixing.

02f2a726fd450b71cb9ad1f23a394d069c47d1f5

In this case, point 4 can be excluded, and indeed any interaction in the modal window is excluded in addition to forms.

Although although this is probably the original idea

plamenradev:
Hello, is point 2. fixed?

_"**When we use asynchronous data loading in the modal window and the screen is updated, the asynchronous data was not requested again.**"_

When the screen is refreshed after failed validation in the modal - the modal stays opened, which is perfect. But when I close the modal and try to open again - it doesn't make the async request to refresh the data in the modal, it just open "the old" failed modal.

PR #2126, opened 2022-02-11
Ability to override default view

G3z:
It is possible, in user-defined screen, to render a different view than default


src/Screen/Screen.php
```diff
@@ -29,6 +29,12 @@ abstract class Screen extends Controller
      * Example: dashboard/my-screen/{method?}
      */
     private const COUNT_ROUTE_VARIABLES = 1;
+    
+    /**
+     * The view that should be rendered
+     *
+     */
+    protected $view = 'platform::layouts.base';
```
tabuna, line 37: Let's use a different method to get the template because a property with such a famous name might already be taken. For example method `$this->screenBaseView()`

PR #2161, opened 2022-03-02
Add placeholder in numberRange for better visibility

rajabilal555:
Before:
image

After:
image


tabuna:
Let's make these strings available for translation into other languages

PR #2167, opened 2022-03-04
custom path for files

### LINKED ISSUE #2166: Custom generator path and disable replication features
### regarding storage path 
in Orchid as far as I have seen if you wan to change the storage path you ll have to extend the engine then change the default engine in the platform config file but that changes the whole app engine , i couldn't just extend the Generator class since the path is dynamic and related to a parent model ( relationship )  and that made me change the File class too , so I had two options :
-  i recreate the Generator and file class and change the constructor to add a path variable
or
- i extend the Generator and File class and add setters before load() .

### regarding attachment model replication
when uploading files, a hash is created for each file , if the same file is uploaded again the model will get replicated without reuploading the files, this made me face an issue when I wanted model files to be in separate manually accessible folders (each model has its own space).
plus if you have old database record of the file you cannot reupload it even if its deleted physically 

## solutions proposed
- add path variable to specify a custom path that will be used instead of Y/m/d 
- add a replicate boolean that will specify if the model should be replicated or not ( true by default )

HijenHEK:
Fixes #2166

## Feature
add the ability to the user to choose a different path without extending the generator, benefits : 

- path can be dynamic 
- you don't have to set only one generator

## Changes

add $path param to `File` and `Generator` constructor after disk with null as default value , null will keep it the default way Y/m/d

```PHP
 return $this->path ?? date('Y/m/d', $this->time());
```


src/Attachment/File.php
```diff
@@ -16,6 +16,7 @@
 use Orchid\Platform\Dashboard;
 use Orchid\Platform\Events\ReplicateFileEvent;
 use Orchid\Platform\Events\UploadFileEvent;
+use phpDocumentor\Reflection\Types\Boolean;
```
tabuna, line 19: I don't think we really use it

src/Attachment/File.php
```diff
@@ -54,19 +60,21 @@ class File
      * @param string|null  $disk
      * @param string|null  $group
      */
-    public function __construct(UploadedFile $file, string $disk = null, string $group = null)
+    public function __construct(UploadedFile $file, string $disk = null, string $path = null, string $group = null)
```
tabuna, line 63: If we add this argument in the middle, then we break backward compatibility for users. 
Can we do it as a method?

```php
$upload = (new File($file, $this->disk))
    ->path('custom/path/etc')
    ->load();
```

src/Attachment/File.php
```diff
@@ -54,19 +60,21 @@ class File
...
     {
         abort_if($file->getSize() === false, 415, 'File failed to load.');
 
         $this->file = $file;
 
+        $this->replicate = true ;
+        $this->path = $path;
```
tabuna, line 70: I don't really understand why we are doing these properties in the current context

src/Attachment/File.php
HijenHEK: yes you are right ! my vscode missimported it , will remove it .

src/Attachment/File.php
HijenHEK: those are irrelevant to the current context, there has been a mix, sorry for this.

src/Attachment/File.php
HijenHEK: Yes , you have a valid point ,

i can add the path method , but 
- the generator construct has to change anyway ( $path param at the end )
- I will have to create a new generator object in the path method since its first instanciated in the constructor
### __construct() in Generator

```php
public function __construct(UploadedFile $file, string $path = null)
```
### path() in File
```php
/**
     * set a custom Path
     * @return File
     */
    public function path(String $path) {
        $generator = config('platform.attachment.generator', Generator::class);
        $this->engine = new $generator($this->file, $path);
        return $this;
    }
```


to prevent re-instantiating the generator we can add a setter for the path variable within  the Generator class
and then we can do something like this

```php
public function path(String $path) {
                $this->engine->setPath($path)
}
```

HijenHEK:
recreated in this [pull request](https://github.com/orchidsoftware/platform/pull/2167)

PR #2168, opened 2022-03-04
allow duplicate files

### LINKED ISSUE #2166: Custom generator path and disable replication features

HijenHEK:
Fixes #2166

## Feature
add the ability to the user to disable replication
 
## Changes

add $replicate boolean to the load() method under File class , by default its true  and then pass null to attachment if replication is disabled

```PHP
        $attachment = $replicate ? $this->getMatchesHash() : null;
```

## usage 
```PHP
        $upload = (new File($file, $this->disk))->load(false);
```


tabuna:
Hi @HijenHEK Please excuse me for not replying for so long. I was in a difficult emotional situation. I think it's a cool suggestion, but what do you think about using it as a method instead of a parameter?

```php
$upload = (new File($file, $this->disk))
    ->allowDuplicates()
    ->load();
```

HijenHEK:
Hello @tabuna ,
I appreciate your efforts! 
yes, i am already implementing a method locally , I just thought of it as a parameter to write less code.
but you have a legit point, we shouldn't change the constructor!

i committed the changes , what I did :

1. added a duplicate property ( works opposite of the replicatre variable )
2. added the allowDuplicates method , it return $this to keep the chain up
3. updated the tests

PR #2170, opened 2022-03-05
fix matrix add row method

KupperXD:
Fixes #


tabuna:
Hi @KupperXD, could you please tell me how to reproduce the bug you are fixing?

KupperXD:
> Hi @KupperXD, could you please tell me how to reproduce the bug you are fixing?

Hi 
In order to recreate the bug described, it's needed  to add a column into matrix, which has a template.
for example, for me it's like this 
I'm creating a Matrix and add an area   Upload::make (standard Orchid)
It consists of  template - 
And that's what matrix_controller.js  is looking for, but it should find a template for a current line coping.

So, in case we add a field to matrix, which has an attached template. so the matrix_controller.js is not going to work properly while row adding.

AlexSabur:
Maybe use stimulus targets instead of `querySelector`?
https://stimulus.hotwired.dev/reference/targets

Rogue85:
@tabuna Could you please merge that?

PR #2236, opened 2022-05-11
Add role Fixed : UserEditScreen fixes #2235

### LINKED ISSUE #2235: Cannot Add a role to a user ( new or existing user )
**Description**
Orchid provides roles to group permissions and adds them to a certain user.
the issue is I cannot add a role to an existing user or create a new user with a role.

**To Reproduce**

  1. got to the users list
  2. select a user to edit ( or click new user )
  3. add a role to it

**Expected behavior**
a role should be assigned to the user 

**Screenshots**
Issue face when trying to update:

image

Issue faced when creating a new user with a role: 

image

HijenHEK:
Fixes #2235

## Issues

when adding roles to a user, the roles array and all other fields are added to the `$userData`  array at the `save()`  method under `UserEditScreen.php` ,
```
$userData = $request->get('user');
```
then the fill method fills the user as if roles is a field in the `users` table and that's what causes these issues, 

instead of that roles should be attached after the fill because they unlike permissions a relationship
```
$user->replaceRoles($request->input('user.roles'));
```

## Proposed Changes

  - removing "roles" key from userData array before filling the user model .

```PHP
       // remove roles array from userData to avoid the fill
       
        unset($userData["roles"]);

        $user
            ->fill($userData)
            ->fill([
                'permissions' => $permissions,
            ])
            ->save(); 
```

### DIFF
--- a/stubs/app/Orchid/Screens/User/UserEditScreen.php
+++ b/stubs/app/Orchid/Screens/User/UserEditScreen.php
@@ -184,6 +184,9 @@ class UserEditScreen extends Screen
             $userData['password'] = Hash::make($userData['password']);
         }
 
+        // remove roles array from userData to avoid the fill
+        unset($userData["roles"]);
+
         $user
             ->fill($userData)
             ->fill([


tabuna:
I wrote it a little differently. Should work with guarded properties since version 12.4.4

PR #2238, opened 2022-05-12
add space between commands

swarakaka:
While it is more than a command and the command buttons have a background, space will be created between them.


tabuna:
Hi @swara-mohammed Thank you very much for the PR, but I'm not sure if I really want this. 
This will cause the phones to shift:

image


If I understand correctly, you want to color the buttons in different colors in this area and therefore decided to add this?

swara-mohammed:
Exactly, that's what I meant, that the buttons can be colored.

PR #2253, opened 2022-05-27
Add group filter to get permissions

astersnake:
## Proposed Changes

  - Currentily if you want to render permissions or create users programattically is all or nothing. This has the intension to return permissions by groups so assign and show permissions is more flexibble according the context.


astersnake:
@tabuna Yes you can use roles, but I think it is a good addition to be able to obtain permissions by group. It also allows the flexibility of not depending on roles if you only want to use permissions.

PR #2259, opened 2022-06-03
fix matrix rows with picture field

### LINKED ISSUE #2258: Matrix with Picture filed bug
**Describe the bug**
Matrix duplicates images for new fields

**To Reproduce**

1. Create Matrix with Picture field. 
 Matrix::make('test')
                ->columns([
                    'Picture' => 'picture',
                    'Text' => 'text',
                ])
                ->fields([
                    'picture' => Picture::make('picture')
                        ->maxFileSize(2)
                        ->targetId(),
                    'text' => Input::make()
                        ->type('text'),
                ])

2. Add picture
3. Save to database
4. Refresh page
5. Try to add new row in matrix

**Expected behavior**
Clear new field without image

**Screenshots**
Screenshot of a bug:
image

This happens because the matrix doesn't clear data-picture-url in template
image

Kostyavas:
Fixes #2258

### DIFF
--- a/resources/views/partials/fields/matrixRow.blade.php
+++ b/resources/views/partials/fields/matrixRow.blade.php
@@ -8,6 +8,7 @@
                     ->prefix($name)
                     ->id("$idPrefix-$key-$column")
                     ->name($keyValue ? $column : "[$key][$column]")
+                    ->set('url', '')
             !!}
         </th>
 


NikitchenkoSergey:
@tabuna check and merge please :)

tabuna:
Guys, I see it, thanks. But I think that we need to solve it differently, so that the state of the fields is always the way it was supposed to be. Without writing the zeroing of each parameter for all types of fields.

As a temporary solution, you can publish this template until a fix is released.

PR #2279, opened 2022-06-18
Wildcard selector for listener.

HichemBenali:
## Proposed Changes
  - Added wildcard to the listener allowing the usage of Listener with Matrix.


tabuna:
Hi @HichemBenali, But after all, this will only work for matrix rows that already existed, but were not added, due to the fact that the event is only registered when the page is loaded?

HichemBenali:
Hi @tabuna you are correct, I forgot that aspect. I believe this can be solved by either attaching a MutationObserver to the matrix table or by adding a custom event to the matrix it self then listening for that event.

HichemBenali:
## Updated Proposal.
- Added wildcard to the listener allowing the usage of Listener with Matrix.
- Added MutationObserver to automatically add event listeners to newly added matrix fields.

PR #2299, opened 2022-07-06
replace diffAssoc with diff so the method is removed when there are parameters in front of it

### LINKED ISSUE #2216: Method save return 404 when there is a parameter id in the url
**Describe the bug**
When we have a parameter in the url like an idenifiant for example a route like this 
https://www.laravel.local/admin/accounts/1/company/add

i got error 404 when i tried to save

**To Reproduce**

  1. Create a route with id parent ( ex : https://www.laravel.local/admin/accounts/1/company/add ) 
  
**Expected behavior**
I have two tables, account and companies, in the account area, we can modify the account information and add companies, I have a 404 error just after I try to add a company 

**Screenshots**
https://ibb.co/5Y8Gfdb
https://ibb.co/1XnQzfK

TomNealBorneman:
Fixes #2216

### DIFF
--- a/src/Screen/Screen.php
+++ b/src/Screen/Screen.php
@@ -216,7 +216,7 @@ abstract class Screen extends Controller
 
         $prepare = collect($parameters)
             ->merge($request->query())
-            ->diffAssoc($method)
+            ->diff($method)
             ->all();
 
         return $this->callMethod($method, $prepare) ?? back();


tabuna:
Hi @TomNealBorneman, could you add some tests to test this case?

TomNealBorneman:
When using Model bindings in your screen method (say the save function for example) you get a 404 when trying to save a new instance of a child relation:

image

image

This happens because it tries to use the method as a key for finding the model. And a model with the key 'save' does not exist. Changing the diffAssoc to diff makes it so the method is removed from the parameters regardless of its position within the collection. The following dumps (made after src\Screen\Screen.php:220) show the different results.

diff:
image

diffAssoc:
image

This still does not work when you use extra url parts like `http://laravel-cms.test/dashboard/users/38/test/create` because then it will still use create as a key for the model binding. But in its current iteration the binding does not work at all.

This fix could cause issues if you have models with string keys that could overlap with methods, say if you have a model with the key `save` it would bind that instead of creating a new model as intended. But this seems like an edge case to me. 

This has no effect on saving existing models like so: `http://laravel-cms.test/dashboard/users/38/test/1` because the identifier is present and can be used for the binding.

caposaric:
Is there anything holding this back from being merged? Seems like a pretty basic requirement to me. 🤞

madyanmalfi:
Changing to diff fixes the issue for me :). If this isn't going to get merged, then please provide anther solution.

PR #2339, opened 2022-08-19
Update ScreenDependencyResolver.php

picko1990:
Resolved 404 error issue that happens when you use laravel `explicit binding`: https://laravel.com/docs/9.x/routing#explicit-binding and use `Implicit Binding` on the `query` function of an orchid screen at the same time


tabuna:
Hi @picko1990 thanks for doing the work, but could you add a simple test?

waotaa:
I ran into the same issue

https://github.com/orchidsoftware/platform/issues/2472

PR #2360, opened 2022-09-10
Add access denied error code customization

Cregennan:
Some applications require not to expose its routes with `403` error code.
If access denied code is e.g. `404` it becomes harder to extract links from application.

### DIFF
--- a/src/Platform/Http/Middleware/Access.php
+++ b/src/Platform/Http/Middleware/Access.php
@@ -56,7 +56,7 @@ class Access
 
         // The current user is already signed in.
         // It means that he does not have the privileges to view.
-        abort(403);
+        abort(config('platform.status', 403));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


Cregennan:
codeclimate doesn't see helper functions

src/Platform/Http/Middleware/Access.php
```diff
@@ -56,7 +56,7 @@ public function handle(Request $request, Closure $next, string $permission = 'pl
 
         // The current user is already signed in.
         // It means that he does not have the privileges to view.
-        abort(403);
+        abort(config('platform.status', 403));
```
tabuna, line 59: Wouldn't it be enough to specify the rule only here? After all, if you want to hide the very presence of limited technologies for example for bots, then this should weed them out.

Cregennan: There can be a case when you want to hide sections of platform from users with lower access level

PR #2396, opened 2022-10-09
getOldValue return float as string

grigoriy-ivanov:
A long number line doesn’t convert to scientific notation

### DIFF
--- a/src/Screen/Field.php
+++ b/src/Screen/Field.php
@@ -316,7 +316,7 @@ class Field implements Fieldable, Htmlable
     {
         $value = old($this->getOldName());
 
-        return is_numeric($value)
+        return is_numeric($value) && is_int($value + 0)
             ? $value + 0
             : $value;
     }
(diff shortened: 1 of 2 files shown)


tabuna:
Hi @grigoriy-ivanov My understanding is that the problem is that large numbers are converted to a string representation, which is not expected. But this PR captures more than what is described. It just disables the float value conversion. 

I think we can solve this as follows:

```php
/**
 * Get the old value of the field.
 * 
 * @return float|int|mixed|string
 */
public function getOldValue()
{
    $value = old($this->getOldName());

    return is_numeric($value) && is_numeric($value + 0)
        ? $value + 0
        : $value;
}
```

and test:


```php
public function testLongNumericValue(): void
{
    Session::start();

    Session::put('_old_input', [
        'numeric' => "1234567890123456789012345678901234567890",
    ]);

    request()->setLaravelSession(session());

    $input = Input::make('numeric')->getOldValue();

    $this->assertEquals('1234567890123456789012345678901234567890', $input);
}
```

grigoriy-ivanov:
Hi @tabuna! It is a very good solution, but my case does not use such large numbers:
```php
        $value = '66666666666666666666';

        $value = is_numeric($value) && is_numeric($value + 0)
            ? $value + 0
            : $value;

        dd($value); //6.6666666666667E+19
```

Issue #2411, opened 2022-10-22
Send non-target extra variables to Listener's asyncMethod


Rfferrao87:
Proposal
---------

I wanted to propose the definition of an `extraVars` array that would send these variables together with the `targets` array in the `Listener` and `listener_controller`, so that the fields specified in the `extraVars` array would not be listened to, but merged with the `targets` array and sent back to preserve the original `query` parameters (while updating the respective `targets` field that triggered the `asyncMethod`).

Example
--------
Let's say I define the `targets` array in `SomeListener` like so
``` lang-php
class SomeListener extends Listener
{
    protected $targets = ['input3'];

    protected $asyncMethod = 'someAsyncMethod';

    ...
}
```
Problem is I have 3 inputs grouped in the same `Layout` so if I change `input3` (triggering the `Listener`) after filling in `input1` and `input2` both of them would be reset because I didn't listen all 3 of them.
``` lang-php
// App\Orchid\Screens\SomeScreen
class SomeScreen extends Screen
{
    ...

    public function someAsyncMethod(string $input1 = null, string $input2 = null, string $input3 = null)  {
        return [
            'input1' => $input1, // null
            'input2' => $input2, // null
            'input3' => $input3, // 'some string 3'
        ];
    }
}

// App\Orchid\Layouts\SomeLayout
class SomeLayout extends Layout
{
    ...

    protected function fields(): iterable
    {
        return [
            Input::make('input1'),
            Input::make('input2'),
            Input::make('input3'),
        ];
    }
}
```
Now using the proposed logic, I'd set the `extraVars` array and set both the `input1` and `input2` with it
``` lang-php
class SomeListener extends Listener
{
    protected $extraVars = ['input1', 'input2'];
    protected $targets = ['input3'];

    protected $asyncMethod = 'someAsyncMethod';

    ...
}
```
With that `input3` remains being the only target for the `Listener` to trigger and the fields specified in `extraVars` are reassembled in the `asyncMethod`

``` lang-php
class SomeScreen extends Screen
{
    ...

    public function someAsyncMethod(string $input1 = null, string $input2 = null, string $input3 = null)  {
        return [
            'input1' => $input1, // 'some string 1'
            'input2' => $input2, // 'some string 2'
            'input3' => $input3, // 'some string 3'
        ];
    }
}
```

plamenradev:
Does $extraVars work with Matrix field? Seems it doesn't join the matrix value at all?
Thanks!

[...]

PR #2412, opened 2022-10-22
Send non-target extra variables to Listener's asyncMethod

Rfferrao87:
#2411

## Proposed Changes

  - Created `extraVars` array in `Orchid\Screen\Layouts\Listener` and added it to `variables` in `build` method
  - Edited `listener.stub` to generate the `extraVars` array when executing `orchid:listener` command
  - Added `extraVars` as `data-listener-extra-vars` in the `listener.blade.php` template
  - Created `get extraVars` method in `listener_controller.js` for merging `extraVars` and `targets` arrays in `render` method
  - Recompiled modified package assets


Rfferrao87:
@tabuna Sorry i'd just like to point out a minor thing that I missed: I should have renamed the closure's parameter in `$this->variables['extraVars'] = collect($this->extraVars)->map(function ($target) {` (in `Orchid\Screen\Layouts\Listener`, line 71) from `target` to `extraVar`, but I didn't notice it at the time; if you'd like I can add another commit to take care of it.

PR #2514, opened 2023-02-03
Register module js

methos02:
Fixes # Register scripts is not work with vite.js and script type module

## Proposed Changes
- register a new key from platform.ressource call module
- generate script tag with type equal to module in app.blade.php


tabuna:
Hi @methos02 Thank you for submitting a PR.
 I'm not using Vite yet, but can't it generate ready-to-use `app.js` that could be specified in the configuration file?

methos02:
I don't think so, we have to use @vite blade methode to generate url with the hash. It's not a big changement, and i can remove the module part to let only vite support.

It's laravel recomandation (and it's very usefull ^^) so it could be great to allow people who want customise js to use it.

PR #2534, opened 2023-02-16
disable read all button if it has no unread messages

swarakaka:


src/Platform/Http/Screens/NotificationScreen.php
```diff
@@ -56,6 +61,9 @@ public function query(Request $request): iterable
             ->paginate(10);
 
         $this->isNotEmpty = $notifications->isNotEmpty();
+        
+        $unreadNotifications = $request->user()->unreadNotifications;
+        $this->hasUnread = boolval(count($unreadNotifications));
```
tabuna, line 66: I think we can write it down better as:

```php
$request->user()->unreadNotifications()->exists()
```
But it still doesn't take type into account. Because there can be different types of notification in the application. And also use the automatic filling of public properties.

I think it should look something like this:
```php
/**
 * @var bool
 */
public $isNotEmpty = false;

/**
 * @var bool
 */
public $hasUnread = false;

/**
 * Query data.
 *
 * @return array
 */
public function query(Request $request): iterable
{
    /** @var \Illuminate\Pagination\LengthAwarePaginator $notifications */
    $prepare = $request->user()
        ->notifications()
        ->where('type', DashboardMessage::class);

    $notifications = $prepare->paginate(10);

    return [
        'notifications' => $notifications,
        'isNotEmpty'    => $notifications->total() > 0,
        'hasUnread'     => $prepare->unread()->exists(),
    ];
}

/**
 * Button commands.
 *
 * @return Action[]
 */
public function commandBar(): iterable
{
    return [
        Button::make(__('Remove All'))
            ->icon('trash')
            ->method('removeAll')
            ->confirm(__('After deleting notifications, this action cannot be undone and all associated data will be permanently lost.'))
            ->disabled(!$this->isNotEmpty),

        Button::make(__('Mark All As Read'))
            ->icon('eye')
            ->method('markAllAsRead')
            ->disabled(!$this->hasUnread),
    ];
}
```

What do you think about it?

swarakaka: However, the values of the $isNotEmpty and $hasUnread variables do not change within the query, that is, they always return the base value and other variables are created. So both buttons will always remain disabled.

solution:
```
  public function query(Request $request): iterable
    {
        /** @var \Illuminate\Pagination\LengthAwarePaginator $notifications */
        $prepare = $request->user()
            ->notifications()
            ->where('type', DashboardMessage::class);

        $notifications = $prepare->paginate(10);
        $this->isNotEmpty   = $notifications->total() > 0;
        $this->hasUnread   = $prepare->unread()->exists();

        return [
            'notifications' => $notifications,
        ];
    }
```

PR #2547, opened 2023-03-02
Add allowAdd() method to the Relation field.

ega22a:
I added the ``allowAdd()`` method to the Relation field.


src/Screen/Fields/Relation.php
```diff
@@ -49,6 +50,7 @@ class Relation extends Field
         'relationSearchColumns' => null,
         'chunk'                 => 10,
         'allowEmpty'            => '',
+        'allowAdd'              => true,
```
tabuna, line 53: Is there any reason why this should be enabled by default?

ega22a: There is no reason... Fixed it, didn't check it right away)

PR #2594, opened 2023-04-25
feat: allow cascaded from Enum

swarakaka:
1.  If the field was cascaded with an Enum then retrieve the value from the object.
2. Using Laravel @class and @checked directive


swarakaka:
Hello @tabuna,
what do you think about this PR?

tabuna:
Can you add a simple test to make sure we don't break support in the future?

PR #2628, opened 2023-05-30
Redundant Argument

kadirermantr:
Hello 👋 

This PR contains a minor correction.

## Proposed Changes

  - The default value of the `domain` variable in the **platform.php** file is already `null`, so there is no need to write it.

image


tabuna:
Hi, I think this is a good change.

PR #2632, opened 2023-06-05
Fix top margin in auth.blade.php

grigoriy-ivanov:
Before
image
After
image


drynyar:
As tmp fix you can create a copy of `vendor/orchid/platform/resources/views/auth.blade.php` template in `resources/views/vendor/platform/auth.blade.php` without `mt-md-auto` classname

https://orchid.software/en/docs/configuration/#overriding-blade-templates

PR #2646, opened 2023-06-16
Fixed mobile menu when not open

d4m:
## Proposed Changes

  - Added visibility: hidden style to .sub-menu in mobile menu when not open


tabuna:
Hi @d4m, can you share the details? Because it works correctly for me when hidden and shown:

| Hidden   	| Show 	|
|---	|:---------:	|
| image   	| image      	|

d4m:
Hi, when the menu is hidden, the submenu links are invisible but still active and clickable.
| Shown  | Hidden |
| ------------- | ------------- |
|  |  |

PR #2654, opened 2023-06-21
Fixes mass assigment on Attachment update

plastique:


tabuna:
Hi, thanks for the PR. Could you give more details? 
By default, the attachment model does not contain fillable `id`

https://github.com/orchidsoftware/platform/blob/1d232452289950366322e77a9d079cebc305a906/src/Attachment/Models/Attachment.php#L32-L46

plastique:
Perhaps this change is not so important.
An error occurs when changing the description or `alt` text of an image, provided that the `preventSilentlyDiscardingAttributes` model setting is `true`.

PR #2656, opened 2023-06-21
Add model description

plastique:
Fixes #

## Proposed Changes

  -
  -
  -


tabuna:
Hey @plastique, thank you for the PR. I think there might be an issue with this proposal for those who modify the models. For instance, I don't have an `email` property for the user, so I would end up seeing it in the IDE suggestions. Do you have any ideas on how we could solve this?

PR #2681, opened 2023-07-24
Feature/translation ja

shibomb:
Add: Japanese Translation file.


tabuna:
Hi @shibomb! Unfortunately, the tests to identify redundant transfers did fail.
For instance, there is a phrase that will never be used: 'Add child'.

shibomb:
Thank you for your time.

PR #2699, opened 2023-08-24
Select from enum feature

noith:
## Proposed Changes

  - Select from enum feature. Supports both backed enum and not backed. Throws exception if provided not enum. Render the right value


tabuna:
Hi @noith, thanks for your contribution. Can you please tell me if you can add tests?

noith:
I did one for not backed enum, didn't find backed in any package

PR #2797, opened 2024-01-28
Add ability to remove input text in Select and Relation fields

grigoriy-ivanov:
### Description:
This pull request adds a new feature to improve the user experience when searching within lists, particularly in the context of multiple selects. When a user selects an item from the search results, a portion of the search phrase remains, which may need to be removed to search for the next item.

image

This feature introduces an option to clear the search phrase automatically after an item is selected. This enhancement streamlines the process of selecting multiple items from a list by reducing the manual effort required to clear the search phrase between selections.

Link to related issue: https://github.com/orchidjs/tom-select/issues/543

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 14.21.0 - 2024-03-19
+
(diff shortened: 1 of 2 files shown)


grigoriy-ivanov:
@tabuna Hi!

tabuna:
@grigoriy-ivanov 

Thank you for bringing this to attention. I believe we should take the opposite approach and implement a method that cancels field clearing instead of enabling it. In my opinion and experience, enabling clearing will likely be required more frequently.

Would it be feasible to make this behavior the default without an option? What are your thoughts on this? Could you update the pull request accordingly?

PR #2815, opened 2024-03-21
Clear input after select item in relation or select fields

grigoriy-ivanov:


grigoriy-ivanov:
HI! @tabuna

I closed #2797 

Now input field is cleared after selecting an item.

PR #2824, opened 2024-04-17
Added a possibility to change 'Add row' label for matrix field

Ercogx:
## Proposed Changes

Added a possibility to change 'Add row' label for matrix field.
Small feature, just to avoid publish view for one line

### DIFF
--- a/src/Screen/Fields/Matrix.php
+++ b/src/Screen/Fields/Matrix.php
@@ -33,7 +33,7 @@ class Matrix extends Field
         'maxRows'           => 0,
         'keyValue'          => false,
         'fields'            => [],
-        'addRowLabel'       => null,
+        'addRowLabel'       => 'Add row',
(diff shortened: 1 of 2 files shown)


src/Screen/Fields/Matrix.php
```diff
@@ -33,6 +33,7 @@ class Matrix extends Field
         'maxRows'           => 0,
         'keyValue'          => false,
         'fields'            => [],
+        'addRowLabel'       => null,
```
tabuna, line 36: It seems that historically we have established a default translation pattern, can you add this? For example: https://github.com/orchidsoftware/platform/blob/5fb66a26ad6f92ec2098e18e57f583626ead262d/src/Screen/Fields/Cropper.php#L63"

Ercogx: Done

PR #2835, opened 2024-05-22
Add frontend controller in screen.

DarKsandr:
Added the ability to assign a stimulus controller to the screen.
Usage:
```
class PlatformScreen extends Screen
{
    ...

    public function frontendController(): string
    {
        return 'hello';
    }

    ...
}
```

### DIFF
--- a/resources/views/app.blade.php
+++ b/resources/views/app.blade.php
@@ -57,7 +57,7 @@
     <div class="row justify-content-center d-md-flex h-100">
         @yield('aside')
 
-        <div class="col-xxl col-xl-9 col-12" @if(!empty($controller))data-controller="{{$controller}}"@endif>
+        <div class="col-xxl col-xl-9 col-12">
             @yield('body')
         </div>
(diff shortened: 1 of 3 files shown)


DarKsandr:
@tabuna

resources/views/app.blade.php
```diff
@@ -57,7 +57,7 @@
     <div class="row justify-content-center d-md-flex h-100">
         @yield('aside')
 
-        <div class="col-xxl col-xl-9 col-12">
+        <div class="col-xxl col-xl-9 col-12" @if(!empty($controller))data-controller="{{$controller}}"@endif>
```
tabuna, line 60: Is this template used and not the one listed as views/layouts ?
/base.blade.php ? it has just the section:


https://github.com/orchidsoftware/platform/blob/3815c39e788bc93c25ccbe8a26ad6b5047715647/resources/views/layouts/base.blade.php#L5

PR #2838, opened 2024-05-29
Visual addition in Boolean component

ega22a:
Fixes
 - Added Bootstrap class `me-1` in variable $class for some pretty view.


tabuna:
Thank you for addressing this issue and proposing a solution. Could you also update the relevant tests to ensure everything works correctly after the changes?

PR #2842, opened 2024-06-13
Add support for modal with XL size and fix $type position

yuweiweiouo:
**Feature:** Added support for modal with XL (Extra large) size.
- Updated modal component to handle XL size configurations.
- Ensured consistent styling and behavior for the new size option.

**Bug Fix:** Corrected the misplaced `$type` variable.
- Moved `$type` to the appropriate location in the layout template.
- Verified that the fix does not impact other functionalities or cause regressions.

### DIFF
--- a/resources/views/layouts/modal.blade.php
+++ b/resources/views/layouts/modal.blade.php
@@ -1,5 +1,5 @@
 @push('modals-container')
-    <div class="modal fade center-scale"
+    <div class="modal fade center-scale {{$type}}"
          id="screen-modal-{{$key}}"
          role="dialog"
          aria-labelledby="screen-modal-{{$key}}"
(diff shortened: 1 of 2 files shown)


resources/views/layouts/modal.blade.php
```diff
@@ -1,5 +1,5 @@
 @push('modals-container')
-    <div class="modal fade center-scale"
+    <div class="modal fade center-scale {{$type}}"
```
tabuna, line 2: The Bootstrap documentation states that size class selectors should be specified for the `modal-dialog` element. This can be seen at the following link: [Optional Sizes](https://getbootstrap.com/docs/5.3/components/modal/#optional-sizes).

Is there a reason why this was moved?

yuweiweiouo: Just wanted to clarify that the `$type` variable is meant for `slide-right`, not for `modal-size`.

I only found the combination `.modal.fade.slide-right` in the CSS, which suggests it's in the wrong spot. It actually works in the modal fade section.

So, I think it makes sense to move it there. Thanks!

yuweiweiouo: you can refer to here [`slide-right`](https://github.com/orchidsoftware/platform/blob/b1125e14e88189915eeb8a3532d7c318b82490df/src/Screen/Layouts/Modal.php#L33)

tabuna: Yes, I definitely made a mistake in my reasoning. 
Thank you for pointing that out. Your patch is already in the release.

PR #2863, opened 2024-07-23
name from presenter class for select field

pLavrenov:
Fixes # RelationController.php


tabuna:
Could you add a test?
It seems that this change makes having a Presentor mandatory for the model.

pLavrenov:
Yes, my mistake, need only check for the presence of the field.

Added a change to the code and now you can set the name via getAttribute() in Model

For example:
`
    public function presenter(): CustomerPresenter
    {
        return new CustomerPresenter($this);
    }

    public function getFullNameAttribute(): string
    {
        return $this->presenter()->fullName();
    }

`

PR #2869, opened 2024-08-05
Allow values in deeper values

mbol8309:
Fixes #
When values comes with "pivot" from pivot tables. It can't be used in form values. 
## Proposed Changes

  - Now with this fix, you can put in fields "pivot.url" and read values from 
`"pivot" : [
"url" => "someurl"
]`
  -
  -

### DIFF
--- a/resources/views/partials/fields/matrixRow.blade.php
+++ b/resources/views/partials/fields/matrixRow.blade.php
@@ -2,9 +2,17 @@
     @foreach($columns as $column)
 
         <th class="p-0 align-middle">
+            @php
+                $columnsArray = explode('.', $column);
+                $value= $row;
+                foreach($columnsArray as $curr) {
+                $value = $value[$curr] ?? "";
+                    if ($value == "") break;
+                }
+            @endphp
             {!!
                $fields[$column]
-                    ->value($row[$column] ?? '')
+                    ->value($value)
                     ->prefix($name)
                     ->id("$idPrefix-$key-$column")
                     ->name($keyValue ? $column : "[$key][$column]")


Fayozjon (approved):
♥

tabuna:
Hi! Thank you for your PR. I’m not entirely clear on the specific result it addresses. Could you please provide a simple example, like:

```php
public function query() {
    return [
        'field' => // ...
    ];
}

public function layout() {
    return [
        Layout::rows([
            Matrix::make('field'),  // ...
        ])
    ];
}
```

This would greatly help me understand the issue better and assist in adding the appropriate test. Thanks!

mbol8309:
Hi,

Sorry for the brief initial post. Let me explain in more detail, and please excuse any language mistakes as English is not my first language.

I am using the CRUD package, which speeds up the creation of some pages, yet it uses the same field attributes from Orchid. When you retrieve an item from a many-to-many relationship on the edit page, and the pivot table contains additional columns, the data is retrieved like this:
```php 
public function query() { 
    return [
        'id' => 5,
        'name' => 'product_x'
        'pivot' => [  //values in pivot table
              'url' => 'some url',
              'another_value' = 'other value'
        ]
    ];
}
```
This is how Laravel returns data in pivot tables. You can read more about it in the Laravel documentation: [Pivot Table Attributes](https://laravel.com/docs/10.x/eloquent-factories#pivot-table-attributes).
Currently, there's no straightforward way to display the values from url and another_value in a field, or at least I haven't found one.

My proposal is to allow field names to include a dot ('.') to enable referencing inner fields. Here's an example of how it could be implemented:
```php 
Matrix::make('providers')->columns([
    'Provider' => 'id',
    'URL' => 'pivot.url'
])->fields([
    'id' => Relation::make('id')
          ->fromModel(Provider::class, 'name')
          ->title('Name')
          ->required(),
    'pivot.url' => Input::make('pivot.url')
                  ->title('Url')
                  ->placeholder('Enter url of product')
                  ->help('Enter url of product in provider domain')
                  ->required()
])
```
I hope this helps, and I believe this code could contribute to the growth of Orchid, which I have recently discovered and already love.

tabuna:
I’ve tried to understand what you are aiming to achieve, but I’m not entirely sure if I’ve grasped your idea completely.

The `Matrix` field is typically used to store various rows of data, each containing specific columns, and then saved in JSON format. For example, in PHP, it might look like this:

```php
[
   ['color' => 'red', 'size' => 12],
   ['color' => 'blue', 'size' => 12]
]
```

So, your structure should resemble the example above. However, I have never tested this functionality with Laravel relationships.

An example of using `Matrix` could be:

```php
// The key name for the query where the value should be stored (e.g., as an array above)
$matrix = Matrix::make('name') 
    ->columns(['color', 'size']);
    
// Note that you don't need to specify names for fields here, as they will be automatically assigned when the row is created
$matrix->fields([
    'color' => Input::make(),
    'size' => Input::make()->type('number'),
]);
```

Therefore, I’m not certain that the code you proposed is entirely correct, as it seems to always set the value only for the second row in the matrix.

I would be happy to clarify any details and help find a solution if I missed something.

mbol8309:
Well, maybe I missed something. But that code worked for me. I used the Matrix fields to represent a many-to-many relationship between two entities where the table containing the foreign keys of the entities also includes additional data corresponding to that relationship. For example, I have products and suppliers, and I modeled this relationship as a many-to-many relationship. But at the same time, I assign additional data to this relationship, such as URL, rating, etc. To add suppliers to the products entity, you need to specify the supplier and the additional data. I solved this using the Matrix field, but I found that when viewing the data, I couldn't see the pivot fields (URL, rating), and the code I uploaded would solve this, as it would work for normal fields and for fields with "." in the name it would look for the values in the returned value.

mbol8309:
In any case, if you believe that this contribution is not aligned with the Orchid platform, feel free to disregard this pull request. I tried to make this contribution because I saw that there were others in the Telegram groups with the same problem who hadn't been able to resolve it.

PR #2880, opened 2024-08-21
Retrieve Screen inherited public properties from parent classes

### LINKED ISSUE #2879: Undefined public property while inherit a Screen class
**Describe the bug**
Public properties inherited by a `Screen` class are not filled with their value if a matching item is found in query method results.

This potentially break inherited logic from the parent class.

**To Reproduce**

  1. Create a class `Parent` that extends Screen with a public property named `$a`;
  2. Define a query method that return an item to assign a value to `$a`;
  3. Create a `Child` class that extends Parent and access `$a` property.

**Expected behavior**
Public properties that have a matching an item from query method should be setted even if they are inherited from the parent class.

**Examples**
Parent.php
```php
<?php

namespace App\Orchid\Screens\Example;

use Orchid\Screen\Screen;

class Parent extends Screen {
    public ?string $a = null;

    /**
     * Query data.
     *
     * @return array
     */
    public function query(): array
    {
        return [
            'a' => 'Example'
        ];
    }
}
```

Child.php
```php
<?php

namespace App\Orchid\Screens\Example;

class Child extends Parent {
    public ?string $b = null;

    /**
     * Query data.
     *
     * @return array
     */
    public function query(): array
    {
        return array_merge(parent::query(), ['b' => 'Another property']);
    }
}
```

Accessing `$b` from child class have a defined value while `$a` is always `null`.

**Server:**
  - Platfrom Version: 14.30.0
  - Laravel Version: 11.x
  - PHP Version:  8.3
  - Database:  N/A
  - Database Version:  N/A

DocLM:
Fixes #2879

## Proposed Changes

  - Iterate over parent classes public properties while creating public properties list;
  - Stop at base Screen class to avoid changes to Orchid native properties;

### DIFF
--- a/src/Screen/Screen.php
+++ b/src/Screen/Screen.php
@@ -266,11 +266,14 @@ abstract class Screen extends Controller
      */
     protected function getPublicPropertyNames(): Collection
     {
-        $reflections = (new \ReflectionClass($this))->getProperties(\ReflectionProperty::IS_PUBLIC);
+        $properties = (new \ReflectionClass(static::class))->getProperties(\ReflectionProperty::IS_PUBLIC);
+        $baseProperties = (new \ReflectionClass(Screen::class))->getProperties(\ReflectionProperty::IS_PUBLIC);
 
-        return collect($reflections)
+        return collect($properties)
+            ->mapWithKeys(fn (\ReflectionProperty $property) => [$property->getName() => $property])
             ->filter(fn (\ReflectionProperty $property) => ! $property->isStatic())
-            ->map(fn (\ReflectionProperty $property) => $property->getName());
+            ->except(array_map(fn (\ReflectionProperty $property) => $property->getName(), $baseProperties))
+            ->keys();
     }
 
     /**


src/Screen/Screen.php
```diff
@@ -266,11 +266,18 @@ protected function fillPublicProperty(Repository $repository): void
...
     protected function getPublicPropertyNames(): Collection
     {
-        $reflections = (new \ReflectionClass($this))->getProperties(\ReflectionProperty::IS_PUBLIC);
-
-        return collect($reflections)
-            ->filter(fn (\ReflectionProperty $property) => ! $property->isStatic())
-            ->map(fn (\ReflectionProperty $property) => $property->getName());
+        $reflection = (new \ReflectionClass($this));
+
+        $properties = new Collection();
```
tabuna, line 271: Do we really need to iterate through all parent classes to find properties?

It seems to me that the following call would yield the same result:

```php
ReflectionClass(static::class)
```

Or am I mistaken?

DocLM: @tabuna You are right, your code will enumerate all public properties without looping.

On the other side it will not protect public properties defined before the original Screen class if not redefined in a sub class but at the moment I dont see properties defined in base classes.

If you agree I will update the PR with your optimized code

tabuna: I reviewed it, and indeed, we are using the public property `$middleware` because we inherit from the controller. I think we can handle the exception in the same way as in the `getAvailableMethods` method. Could you please update the PR?

DocLM: Hi @tabuna,
I made a rebase with latest changes and improved the code with all the suggestions.

PR #2887, opened 2024-08-28
Add AJAX options to Select field

Lucisu:
This PR adds the option to load Select values via AJAX using Tom's own [remote feature](https://tom-select.js.org/examples/remote/).

Example:

```
Select::make( 'cities' )
    ->title( __( 'Cities' ) )
    ->ajaxOptionsUrl( 'https://example.com/api/cities' )
    ->horizontal(),
```

The API must return an array like:

```
{
    "options": [
        {
            "value": 1,
            "label": "Example"
        }
    ]
}
```

I also added the option to change the "value" and "label" key names.

---

I think it's valid to discuss the new attributes' names first and then decide whether to add them as inline attributes or normal ones.

### DIFF
--- a/resources/js/controllers/select_controller.js
+++ b/resources/js/controllers/select_controller.js
@@ -48,8 +48,7 @@ export default class extends ApplicationController {
                 axios
                 .get(ajaxOptionsUrl, { query })
                 .then((response) => {
-                    console.log(1);
-                    callback(response.data.items);
+                    callback(response.data.options);
                 });
             };
         
--- a/resources/js/controllers/select_controller.js
+++ b/resources/js/controllers/select_controller.js
@@ -37,16 +37,18 @@ export default class extends ApplicationController {
             }
         };
 
-        const fromUrl = select.getAttribute('fromUrl');
+        const fromUrl = select.getAttribute('fromurl');
 
         if (fromUrl) {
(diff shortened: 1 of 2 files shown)


tabuna:
I find this to be an useful addition. However, a few things concern me:

We already have the following methods:

```php
Select::make('user')
    ->fromModel(User::class, 'email', 'uuid');

Select::make('user')
    ->fromQuery(User::where('balance', '!=', '0'), 'email');
```

It seems to me that, by analogy, we could add a method like this:

```php
Select::make('user')
    ->fromUrl('https://example.com', 'label', 'value');
```

This would be more consistent and intuitive for the user, making it easier to understand. What do you think?

tabuna:
I’m also concerned that we aren’t handling this the same way as in the `Relation` field, which was originally designed for the same purposes.

https://github.com/orchidsoftware/platform/blob/8a786baa55eb0c9909c70eedff3f9520f5cdbaea/resources/js/controllers/relation_controller.js#L30-L31

It might be a good idea to combine both fields into one, or at least use a consistent approach for now.

Lucisu:
> This would be more consistent and intuitive for the user, making it easier to understand. What do you think?

It makes total sense! I like this approach, and just sent a new commit to change it.

> It might be a good idea to combine both fields into one

I agree that having both might be confusing. Since the Select can also retrieve from a model and query, there may be no reason for the Relation one.

PR #2889, opened 2024-08-31
Prevent CSRF token from being leaked to cross-origin requests

Lucisu:
Currently, the code inside the html_load_controller.js makes the CSRF token being sent to every Axios and Turbo request:

https://github.com/orchidsoftware/platform/blob/7e63b9c68af945c09f1dbb50896ad9f135a0e324/resources/js/controllers/html_load_controller.js#L41-L46

This, however, leaks the token to cross-origin requests as well, defeating the whole purpose of CSRF protection.

This fix prevents the token from being leaked to cross-origin requests.

For Axios, it is certain that it needs to be implemented. For Turbo, I'm not sure the "turbo:before-fetch-request" is triggered during cross-origin requests, but I added the fix there as well, just in case.

### DIFF
--- a/resources/js/controllers/html_load_controller.js
+++ b/resources/js/controllers/html_load_controller.js
@@ -41,8 +41,21 @@ export default class extends ApplicationController {
         window.axios.defaults.headers.common['X-CSRF-TOKEN'] = token.content;
         window.axios.defaults.headers.common['X-Requested-With'] = 'XMLHttpRequest';
 
+        axios.interceptors.request.use(function (config) {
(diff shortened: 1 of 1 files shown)


tabuna:
Hi! Thanks for pointing that out.

To be honest, I'm not sure it's still necessary to manually set the CSRF token this way. Back when Hotwire was still Turbolinks, it was required, but now the framework should automatically fetch the value from:

```html
<meta name="csrf-token" content="[your-value]">
```

Maybe we should try removing this and rely entirely on the framework?

PR #2895, opened 2024-09-10
Feature/filters autofill

Mystical0628:


src/Filters/Autofill.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Orchid\Filters;
+
+trait Autofill
```
tabuna, line 7: I’ve tried several times to think of scenarios where this trait might be unnecessary or could potentially break something, but frankly, I couldn't come up with any.

What do you think—if the value is placed directly in the class, could that cause any issues?

Mystical0628: I think so too. This feature can only break the logic if someone uses this component incorrectly, but I think this cannot happen in a serious production project.

Mystical0628: Fix pull request. Move code from trait to Filter class and fix Tests

PR #2907, opened 2024-10-12
Fix error of decoding base64-string containing Cyrillic characters.

ega22a:
After updating the SimpleMDE field, where a bug with HTML code processing was fixed, a problem with displaying UTF-8 encoded characters appeared. It is related to the fact that JavaScript `atob()` function processes a string only in ASCII.

I wrote a wrapper to restore justice to UTF-8.


tabuna:
Thank you for the PR! However, I think it's better to revert back to using `JSON`, as the initial issue was that the field was getting its value from attributes that are automatically escaped.

```php
// contains the escaped value meant to be inserted into a DOM element
$attributes['value'];


// unescaped value that should be passed to the JS handler
$value;
```

I rushed the release, which unfortunately broke this logic. I apologize for the inconvenience :(

ega22a:
> Thank you for the PR! However, I think it's better to revert back to using `JSON`, as the initial issue was that the field was getting its value from attributes that are automatically escaped.
> 
> ```php
> // contains the escaped value meant to be inserted into a DOM element
> $attributes['value'];
> 
> 
> // unescaped value that should be passed to the JS handler
> $value;
> ```
> 
> I rushed the release, which unfortunately broke this logic. I apologize for the inconvenience :(

As best I can understand, the 14.37.0 release fixed this and my Pull Request can be rejected.

As time permits, I will check the functionality of the release and if everything is fine, I will cancel this Pull Request. Napoli will have any problems, I will report back 🤓.

PR #2935, opened 2025-01-13
Add method labels in Tabs

kalexhaym:
Fixes #

## Proposed Changes

- add new feature in tabs
- new method "labels" that allows you to set custom tab names
- this may be necessary when you need to make several tabs on one page with the same labels but with different keys


tabuna:
Hi! Sorry for the delay. 😊 I'm a bit confused—how would renaming help us properly switch between tabs with the same name?  

Switching happens based on the `id`, not the name. To fix this, we could add a prefix in the template, for example:  

```php
data-bs-target="#tab-{{ \Illuminate\Support\Str::slug($templateSlug . $name) }}"
```  

What do you think? 😊

PR #2939, opened 2025-01-22
Added activeAccordion method to Accordion.php

bald-cat:
Fixes #

## Proposed Changes

  - Added open method to Accordion.php. Allows you to specify which accordions should be open when the page is opened. Accepts a string and an array, if the method is not called, the first accordion will be opened by default, as it was before adding the method

### DIFF
--- a/src/Screen/Layouts/Accordion.php
+++ b/src/Screen/Layouts/Accordion.php
@@ -55,4 +55,18 @@ abstract class Accordion extends Layout
 
         return $this;
     }
+
+    /**
+     * Specifies the accordion that should be opened by default.
+     *
+     * @param string $activeAccordion
+     *
+     * @return $this
+     */
+    public function activeAccordion(string $activeAccordion): self
+    {
+        $this->variables['activeAccordion'] = $activeAccordion;
+
+        return $this;
+    }
 }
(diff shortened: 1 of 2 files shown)


src/Screen/Layouts/Accordion.php
```diff
@@ -57,16 +65,21 @@ public function stayOpen(bool $stayOpen = true): self
...
      */
-    public function activeAccordion(string $activeAccordion): self
+    public function activeAccordion(string|array $activeAccordion): self
     {
-        $this->variables['activeAccordion'] = $activeAccordion;
+        $this->variables['open'] = $this->openSet
+            ? array_merge($this->variables['open'], Arr::wrap($activeAccordion))
+            : $activeAccordion;
+
+        $this->openSet = true;
```
tabuna, line 82: What is the purpose of the `openSet` property in our case?

bald-cat: When creating a class, we set a standard key for the first tab in the constructor to preserve the standard previous behavior.

A flag property is needed so that when the method is first called, it can be distinguished whether it was called earlier and, depending on this, merge the array or overwrite it.

PR #2950, opened 2025-01-31
Added color row for tables

bald-cat:
Fixes #

## Proposed Changes

  - Possibility to choose the color of a row in a table.
  - Added method contextual() in Table.php
  - Added method contextualClass() in Color enum
  - Added call closure function in table.blade.php
  - Default using Color::DEFAULT in contextual method not rewrite


tabuna:
Are you sure you want to do this in 15? Maybe we should do it in master instead?

PR #2972, opened 2025-03-12
Resolve #2971: Improve table header text contrast for accessibility

### LINKED ISSUE #2971: Accessibility Issue: Insufficient Contrast in Table Header Text
**Describe the bug**
The table header text appears in a light gray color (`text-muted`), making it difficult to read against the background, especially for users with low visibility or accessibility challenges. The lack of sufficient contrast causes readability issues.

**To Reproduce**

1. Open the webpage containing the table.
2. Look at the header row of the table.
3. Notice that the text is rendered in a light gray color (`text-muted`), which is not easily visible against the background.

**Expected behavior**
The table header text should have a stronger color contrast with the background to provide better readability.

sahar-saadatmandi-corelab:
Fixes #2971

## Proposed Changes

- Added the `text-dark` class to `<th>` elements to improve the contrast and accessibility of table headers.
- Replaced the `text-muted` class with `text-dark` on `<a>` elements to enhance link visibility and readability.

### DIFF
--- a/resources/views/partials/layouts/th.blade.php
+++ b/resources/views/partials/layouts/th.blade.php
@@ -1,11 +1,11 @@
-<th @empty(!$width) width="{{$width}}" @endempty class="text-{{$align}}" data-column="{{ $slug }}">
+<th @empty(!$width) width="{{$width}}" @endempty class="text-{{$align}} text-dark" data-column="{{ $slug }}">
     <div class="d-inline-flex align-items-center">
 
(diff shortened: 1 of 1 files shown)


tabuna:
Thank you, this is really interesting! However, it seems to introduce a new issue. When sorting was active, the column was highlighted with a darker font, but now everything is dark, making it difficult to tell which column is being sorted. How about using opacity to improve visibility? What do you think?

PR #3009, opened 2025-05-03
Added `Toggle` action button

tabuna:
This PR introduces a new `Toggle` action button that functions as a boolean switch. It acts as a button that reloads the page when triggered, but offers a more familiar and intuitive interface. This action can be particularly useful in scenarios like tables, where a boolean value needs to be toggled with a user-friendly UI.

```php
use Orchid\Screen\Actions\Toggle;

Toggle::make('Enabled')
    ->method('switch', ['user' => $user])
    ->status(true);
```

https://github.com/user-attachments/assets/8eae0c70-4aa9-45f7-a017-731b5e649cdd


tabuna:
Hi @StenSeegel 


This works just like a regular button or link — you can display it, for example, inside a table cell like this:

```php
TD::make()->render(function(User $user) {
    return Toggle::make('Enabled')
        ->method('switch', ['user' => $user])
        ->status($user->active);
});
```

Then, just add the method specified in `method('switch', ...)` to your screen:

```php
public function switch(User $user)
{
    $user->active = ! $user->active;
    $user->save();
}
```

That’s it — now the toggle will update the user’s status when clicked.

PR #3047, opened 2025-10-13
#2955 Hide extra div for hidden inputs

### LINKED ISSUE #2955: Gap when using hidden input
When using hidden input in rows layout additional gap adds around it because this layout has flex container with gap and input has not hidden container.

Reproducing:
```
Input::make('some-data')->hidden(),
Input::make('like-id')->hidden(),
Input::make('or-something')->hidden(),
```

Result (modal example):
Image

Problem goes away when wrapping inputs in group like this:
```
Group::make([
    Input::make('some-data')->hidden(),
    Input::make('like-id')->hidden(),
    Input::make('or-something')->hidden(),
]),
```
or I can just redefine input/layout template, but this behavior still looks like a bug.

alikhanz:
Fixes #2955

## Proposed Changes

  - Hide extra div that gives unneccessary gap in layout.

### DIFF
--- a/resources/views/fields/input.blade.php
+++ b/resources/views/fields/input.blade.php
@@ -1,8 +1,5 @@
 <div data-controller="input"
      data-input-mask="{{$mask ?? ''}}"
-     @class([
-        'd-none' => $attributes['type'] === 'hidden' || $attributes['hidden']
-     ])
 >
     <input {{ $attributes }}>
 </div>
--- a/src/Screen/Fields/Input.php
+++ b/src/Screen/Fields/Input.php
@@ -112,6 +112,10 @@ class Input extends Field
             if (is_array($mask)) {
                 $this->set('mask', json_encode($mask));
             }
+
+            if ($this->get('type') === 'hidden' || $this->get('hidden', false)) {
+                $this->hiddenFormType();
+            }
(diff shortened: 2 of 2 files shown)


resources/views/fields/input.blade.php
```diff
@@ -1,5 +1,8 @@
 <div data-controller="input"
      data-input-mask="{{$mask ?? ''}}"
+     @class([
```
tabuna: I think this is a useful PR 👍
But maybe it would be better to add this directly in the class constructor:

```php
$this->beforeRender(function () {
    if ($this->get('type') === 'hidden' || $this->get('hidden', false)) {
        $this->hiddenFormType();
    }
});
```

This way, we’ll handle it in a more universal and elegant way.

alikhanz: > I think this is a useful PR 👍 But maybe it would be better to add this directly in the class constructor:
> 
> ```
> $this->beforeRender(function () {
>     if ($this->get('type') === 'hidden' || $this->get('hidden', false)) {
>         $this->hiddenFormType();
>     }
> });
> ```
> 
> This way, we’ll handle it in a more universal and elegant way.

lgtm, im updated PR

PR #3051, opened 2025-10-24
ref #3050 Fixed error when nesting groups

### LINKED ISSUE #3050: Is nested groups allowed?

keasy9:
Fixes #3050

Changes:

  - added check if child field is group in group rendering method


tabuna:
You could add a simple unit test to make sure we don’t accidentally break this behavior in the future and to confirm that this behavior is considered valid.

tabuna:
I think it would be a good idea to add the example from the issue to `Orchid\Tests\Unit\Screen\Fields\GroupTest` so we can make sure that a `Group` can be nested inside another `Group`.

PR #3061, opened 2025-11-16
Avoid invalid image request in Firefox caused by the template.

bandicsongor:
Firefox is the only browser that processes the contents of <template> tags and attempts to load the image referenced with src="{url}". This results in a 404 response, and because my server is configured with Fail2Ban to block repeated 404 requests, Firefox ends up banning me from my own server.

This issue occurs only in Firefox—Chrome and other browsers ignore the contents of <template> elements as expected.

**Solution:**
I initialized the template’s image with a 1-pixel transparent placeholder and then set the real URL via a query selector instead of performing a string replace on innerHTML. I also added an alt attribute to the image.

### DIFF
--- a/resources/js/controllers/attach_controller.js
+++ b/resources/js/controllers/attach_controller.js
@@ -185,11 +185,15 @@ export default class extends ApplicationController {
         preview.querySelectorAll('*').forEach(element => {
             preview.innerHTML = preview.innerHTML
                 .replace(/{id}/gi, attachment.id)
-                .replace(/{url}/gi, attachment.url)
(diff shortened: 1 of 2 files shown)


tabuna:
I tried uploading and deleting images in `Attach` both in multiple mode and single mode. I tested it in Firefox 145.0 (aarch64) and I’m not seeing any 404 errors.

image

Could you please explain in more detail how I can reproduce the issue?

bandicsongor:
I’ll try to reproduce the issue in the next couple of days. Please allow me 1–2 days and I’ll get back to you with the results.

PR #3134, opened 2026-09-04
Redraw chart when its modal or accordion pane is shown

### LINKED ISSUE #3089: Multiple Charts brake in modals or accordions.
**Describe the bug**
When displaying charts via modals or via accordions, the charts filled area render breaks and only dots are visible on the graph. Only the first rendered chart works fine.

**To Reproduce**

1. Have different charts show via a modal window or in an accordion component.
2. Open/expand one after another

**Expected behavior**
Only the first chart will have the filled area rendered. All the others will just show the dots.

**Screenshots**

Image

**Desktop (please complete the following information):**

- OS: Ubuntu 24
- Browser Firefox 142

**Server (please complete the following information):**

- Platfrom Version: 14.0
- Laravel Version: 10
- PHP Version: 8.4
- Database: MariaDB
- Database 12.3

wakqasahmed:
The chart is constructed at `new Chart(this.data.get("parent"), ...)` in `connect()`, which sizes itself off the container's width at that moment. Inside a modal or an accordion pane that's still `display:none`, that's zero — the SVG comes out with `width="0"` and garbage path coordinates. The existing `shown.bs.tab` listener already redraws it for the tab case, but nothing covers a modal or a collapse, matching what's reported: only the chart that happens to be visible on load renders correctly.

Same fix as the tab one: find the closest `.modal` / `.collapse` ancestor and redraw on `shown.bs.modal` / `shown.bs.collapse`.

Reproduced this outside the app with the actual `frappe-charts` build and a real Bootstrap 5 modal/collapse — confirmed a chart built while hidden renders with `width="0"` and negative/garbage path coordinates, and that firing the same redraw this PR adds fixes both cases (`width` comes back correct and the path matches a chart built while visible). No JS test harness exists in this repo for Stimulus controllers (same as #3130), so that browser check is what I verified against instead of an automated test.

Fixes #3089

### DIFF
--- a/resources/js/controllers/chart_controller.js
+++ b/resources/js/controllers/chart_controller.js
@@ -75,7 +75,10 @@ export default class extends ApplicationController {
         }
 
         if (this.collapse !== null) {
-            this.collapse.removeEventListener("shown.bs.collapse", this.drawEvent);
+            this.collapse.removeEventListener(
+                "shown.bs.collapse",
+                this.drawEvent
+            );
         }
     }
 }
--- a/resources/js/controllers/chart_controller.js
+++ b/resources/js/controllers/chart_controller.js
@@ -36,6 +36,19 @@ export default class extends ApplicationController {
         document.querySelectorAll('a[data-bs-toggle="tab"]').forEach(tabElm => {
             tabElm.addEventListener("shown.bs.tab", this.drawEvent);
         });
+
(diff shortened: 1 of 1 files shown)


wakqasahmed:
Closing in favor of #3135 — this PR had the wrong commit author (a stray local git config override, not you).

